PR #21, opened 2026-04-03
Add these motor control files to aquilla-main

NicoleCornell8:
these are the new motor control files we have made that we use for drawer and axis motor controls now instead of the ones arete gave us. I hope I did this correctly as in making a new branch and then requesting a pull - Max

PS -- we also put lod_verification_all.py in here. We would like it instead of the other lod_verification.py file (which is not needed)

### DIFF
--- a/lod_verification_all.py
+++ b/lod_verification_all.py
@@ -1,0 +1,65 @@
+import sys
+from adc_class import OpticalRead
+from led_class import LED
+import time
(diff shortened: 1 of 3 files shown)


NicoleCornell8:
Looks, good. Next time just delete the files you don't want as well or replace it with your changes. Will change that for you :)

PR #50, opened 2026-04-22
Adjusted drawer speed by editing motor_class.py

Rabbott314:
!My commit title incorrectly states **drawer**_class.py, the edited file was **motor**_class.py!

Increased drawer speed by lowering the value at line 148 in motor_class.py - in the def open ( self ) function - to be 0.0005 instead of 0.002. Jake previously lowered the value from 0.002 to 0.001 to increase speed, but this change was reverted in 'dockerization.'

### DIFF
--- a/aq_lib/motor_class.py
+++ b/aq_lib/motor_class.py
@@ -145,7 +145,7 @@ class Drawer ( Motor ):
 
     def open( self ):
         self.home()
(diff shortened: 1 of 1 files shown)


NicoleCornell8 (approved):
Looks good, simple fix!

PR #68, opened 2026-05-15
Plz change teh adc limits

Rabbott314:
please change -- device goes past physical limits if not given a max -- but with home at the end now


NicoleCornell8 (approved):
Looks good.

PR #78, opened 2026-05-28
Add files via upload

Rabbott314:
please change the motor axis controls to accept the proper argument from teh config file -- there is no longer a "well one" designation to pull from. Actually uncleaar is this is even the fix, but it is what Claud said:

image

### DIFF
--- a/motor_axis (2).py
+++ b/motor_axis (2).py
@@ -1,0 +1,367 @@
+#!/usr/bin/python3
+"""
+Axis Motor Diagnostic Tool
+
+The axis motor positions the optics over the wells. There are 6 positions:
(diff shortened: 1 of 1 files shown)


Rabbott314:
oh also, we need these two recipes on SN02 (going to Duke), but not necessarily on the other devices. Jake copied them onto SN02 for now, but just so you have them as well
[STEC_and_EPEC_Hygiena1.75_no37Cstep.json](https://github.com/user-attachments/files/28362555/STEC_and_EPEC_Hygiena1.75_no37Cstep.json)

[O157__ESBL_full.json](https://github.com/user-attachments/files/28362488/O157__ESBL_full.json)

PR #121, opened 2026-06-04
Implemented Countdown Timer

### LINKED ISSUE #110: Add Countdown/Timer instead of stopwatch
## Summary

Certain tested profiles have an estimated runtime necessary for completion. If one of these profiles are selected display the estimated runtime necessary as a countdown rather than a stopwatch.

Changes Requested:

- Add additional data in profile bundle json files for selected profiles. (json files, history, profiles.html)
- Check json file for this information, if it is present, use this data as the countdown. Else, use the original countup.
- Edit change profile page to allow user to add a optional estimated run time.
- Ensure that device run is not dictated by the countdown. If run finishes before estimated time, switch to done screen. 
- (Optional) if run reaches 0, add a flag that 

## Motivation

The goal is to improve user quality of life.

## Spec

Will you write a spec before implementing?

- [ ] Yes — I will create `specs/[type]/[feature-name].md` before starting

## Acceptance Criteria

How will we know this is done?

- [ ] Tests pass: `pytest tests unit_tests -v`

jackhu23:
## What Changed

On profile creation or profile edit, user can now specify an optional "Est. Time" for the profile. If there is a time specified, on machine run, a countdown time will be show rather than the current stopwatch.

## Why

https://github.com/AcornGenetics/aquilla-main/issues/110

## Spec

`specs/frontend/countdown-timer.md`

## Changes Made

- [x] Source code modified
- [x] New tests written

## Hardware Testing

- [x] Tested in simulation mode only — reason: [why physical not possible]

## Reviewer Notes


jackhu23:
# Tests Implemented

## Unit Tests:

### `estimated_minutes_to_seconds()`:
-  Ensure minutes successfully being converted to seconds.
- Invalid entries (0, negatives) result in None output

### `_order_time_fields()`: 
- Ensure that `time_unavailable` + `estimated_completion_seconds` are inserted before `steps` in profile json files.
- Ensure existing keys and values are preserved

## Contract Tests:

- Ensure end to end behavior for inputting a valid time or none is either its corresponding seconds or none.
- Ensure that editing a profile with time clears the value and replaces it with null

## E2E DOM Tests:

- Ensures that `finishing run` flag is present
- Ensures that run countdown is present
- Ensures that "optional" is present in est. time field

NicoleCornell8 (approved):
Looks good, will try on sn03 first!

NicoleCornell8 (changes requested):
## 🔴 Blocking: two bundled profiles lost their `optics` step

`profiles/bundled/ryan_thermal_tests.json` and `profiles/bundled/verification_profile.json` both had a top-level `{"optics": ""}` step silently dropped, and the files got fully re-serialized (2-space indent, a `labels` block appended, trailing newline removed). The other four bundled profiles in this PR were edited cleanly — only the two new timer fields added — so these two stand out.

This isn't cosmetic. `optics` is a real run directive the thermal engine consumes:

```python
# aq_lib/thermal_parser.py:41-42
elif "optics" in s:
    yield "optics", n, setpoint, setpoint, duration, last_time
```

So removing it changes the optics-acquisition sequence of those two protocols.

### How it happened (and why it's a bigger deal than this PR)

I'm fairly sure these two files were opened in the Edit Profile UI to type in the new "Est. Time" value, rather than hand-edited. The editor doesn't know what an `optics` step is — `parseProfileStep` in `edit.js` only handles six types and returns `null` for everything else:

```js
const parseProfileStep = (step) => {
  if (step.setpoint  !== undefined) { ... }
  if (step.ramp_rate !== undefined) { ... }
  if (step.enable    !== undefined) { ... }
  if (step.disable   !== undefined) { ... }
  if (step.pcr_fanon !== undefined) { ... }
  if (step.pcr_fanoff!== undefined) { ... }
  return null;          // {"optics": ""} lands here → silently dropped
};
```

`buildStepsPayload` (the save path) only re-emits those same six types, so a load → save round-trip eats the optics step. The evidence lines up perfectly: in `verification_profile.json` only the *standalone* optics step was lost while the one *inside* the `repeat` block survived with its original tab indentation — exactly what you'd expect since the editor handles staged blocks separately.

**For this PR:** please restore the step content of both files and add only `time_unavailable` / `estimated_completion_seconds` by hand, matching how the other four were done. Happy to hand you the corrected versions.

**Separately:** this exposed a latent data-loss bug — the profile editor silently discards any step type it doesn't recognize on every save, so any operator who edits an optics-containing profile in the UI corrupts the protocol with no warning. Worth its own issue; not on you to fix here.

## 🟡 Worth a look: countdown is lost on a mid-run page reload

`cachedEstimateSeconds` is only set by `loadProfileLabels()` on profile selection. If the kiosk reloads or the WS reconnects during a run, `beginTimerMode()` fires with it still `null` and silently falls back to the stopwatch for the rest of the run. The spec already allows a silent stopwatch fallback, so this is arguably fine — but a refresh permanently downgrading the display feels surprising. A sentence in the §4 error-states note, or sourcing the estimate from the running profile the backend reports, would close it. Fine as a follow-up.

## Minor / nits (non-blocking)

- `estimated_minutes_to_seconds` carefully handles floats/NaN/inf, but `ProfileSave.estimated_minutes: Optional[int]` means Pydantic coerces before the helper ever sees a float — so the decimal-rounding path is only reachable from the unit tests, not the live route. Harmless, just noting the real API can't actually send `2.6`.
- `formatRemaining` reuses `lastElapsedSeconds` to hold the *remaining* value, which makes the freeze-on-stop work correctly, but the variable name now means two things depending on mode. Slightly muddy, not wrong.

Everything else — validation on both ends, always writing both keys, the contract test that proves omitting the field preserves an existing estimate, the modal mirroring the stopping-modal pattern — looks solid. Flip the two profiles back and I'm happy to approve.

profiles/bundled/ryan_thermal_tests.json
```diff
@@ -1,94 +1,92 @@
...
+    {
+      "ramp_rate": 1.75
+    },
+    {
+      "pcr_fanon": 1
+    },
+    {
+      "enable": 0,
+      "duration": 1,
+      "description": "Turn on and record temperature for a little bit"
```
NicoleCornell8: 🔴 The standalone `{"optics": ""}` step that used to sit right here (between the `enable` step and `setpoint: 25`) is gone. `optics` is consumed by the thermal engine (`aq_lib/thermal_parser.py:41`), so this changes the run protocol. Looks like a side effect of round-tripping this file through the Edit Profile UI — `parseProfileStep` in `edit.js` returns `null` for `optics` and drops it on save. Please restore the original steps and add only the two timer fields by hand.

profiles/bundled/verification_profile.json
```diff
@@ -1,49 +1,48 @@
...
-        "pcr_fanon" : 1
+      "ramp_rate": 1.6
     },
     {
-        "enable": 0, 
-        "duration": 1,
-        "description": "Turn on and record temperature for a little bit"
+      "pcr_fanon": 1
     },
     {
```
NicoleCornell8, line 24: 🔴 Same issue as `ryan_thermal_tests.json`: this file originally had **two** `{"optics": ""}` steps and the standalone top-level one (between `enable` and `setpoint: 25`) was dropped here — the one inside the `repeat` block survived. Editor round-trip side effect. Please restore the original steps and add only `time_unavailable` / `estimated_completion_seconds` by hand.

NicoleCornell8 (changes requested):
Second review. Countdown logic itself is solid — server-driven (`remaining = estimate − panel.elapsed`), clamps at 00:00, fail-safe stopwatch fallback, tests green (37 passed).

But the blocker from the earlier review is **not fixed** — the "Reverted 2 changed profiles" commit didn't restore them. Verified against `origin/main`:

- `ryan_thermal_tests.json`: standalone `{"optics": ""}` step **still dropped**, and the `repeat`/`cycles:1` block got **unrolled** into flat steps.
- `verification_profile.json`: standalone `optics` step **still dropped** (the one inside `repeat` survived).

Both are still re-serialized (2-space indent, `labels` appended, trailing newline gone). So the optics-acquisition sequence of these two protocols is still altered. Please restore both files from main and add only `time_unavailable` / `estimated_completion_seconds` by hand.

Minor (follow-up, not blocking): on a mid-run reload `cachedEstimateSeconds` is null so it silently drops to stopwatch (already noted); and on stopping a countdown run, `lastElapsedSeconds` holds *remaining* seconds, so the freeze re-renders that value via `formatElapsed` — cosmetic.

NicoleCornell8 (changes requested):
Whoops disregard my harsh comment, all good, I fixed branch so it is up to date when you pull :)

Issue #159, opened 2026-06-15
Build watch_cancel.py — live terminal monitor for self-cancellation


jackhu23:
## What to build

A small stdlib-only terminal app that tails the controller, web-app, and lid-heater logs in real time, colorizes the key fingerprint strings, and shows a live header. Per `specs/analysis/sentri-self-cancel-study.md` §6.2. Run in a second SSH session while reproducing.

Header shows: current lid-worker `live=N` | consecutive poll-failure count `k/10` | seconds since the web app last logged | last cancel verdict. Red banner when `Stop request detected` or `Backend unreachable` appears.

## Triggers addressed

- **Trigger 2 (safety net)** — the `k/10` poll-failure ladder and a "seconds since app log" counter climbing past 5 s expose H2 (event-loop block) / H7 (restart) live.
- **Threading (H1)** — the live `live=N` count rising mid-session shows the leak as it happens.

## Acceptance criteria

- [ ] Follows all three logs concurrently without busy-spinning
- [ ] Colorizes: forced-stop (red), poll ladder (yellow), stop/run-stopped (magenta), lid events (cyan), stop-pressed/restart (red)
- [ ] Header live-updates lid `live=N`, poll-failure count, and seconds-since-app-log
- [ ] Red banner fires on `Stop request detected` / `Backend unreachable`
- [ ] Accepts native or Docker log paths via args

## Blocked by

None - can start immediately.

Issue #158, opened 2026-06-15
Build scan_cancel_logs.py — log fingerprint classifier (Trigger 1 vs 2 vs Threading)


jackhu23:
## What to build

A stdlib-only script that parses the controller log + web-app log, finds each self-cancel, and **classifies it** against hypotheses H1–H8 (or H0/UNKNOWN), plus the timestamp **co-incidence delta** (did the web app go silent at the instant the controller gave up?). Per `specs/analysis/sentri-self-cancel-study.md` §6.1, with unit tests from §10.

Detects the code-grounded fingerprints: `Backend unreachable … forcing stop`, the `Error polling stop request (k/10)` ladder, `Stop request detected`, `Run stopped by user`, controller-wide comms errors, `LID WORKER … live=N`, `LID JOIN DONE`, and (app log) `Stop button pressed` / `Stop reset` / `Application startup`.

## Triggers addressed

This is the **master discriminator** across all three buckets:
- **Trigger 1 (real flag)** — `Stop button pressed` with no tap → H3 (phantom touch / frontend re-fire); flag true with no `Stop reset` → H8 (stale flag).
- **Trigger 2 (safety net)** — `Backend unreachable` + no `Stop button pressed` → H2/H4/H5/H6/H7 (returns the candidate set; narrow with snapshot_resources).
- **Threading (H1)** — lid `live=N` peak > 1 in the window.
- **H0 / UNKNOWN** — no fingerprint matched; never asserts a cause it can't prove.

## Acceptance criteria

- [ ] Finds each self-cancel window and slices both logs by timestamp around it
- [ ] Returns the correct verdict (or candidate set) per the §3 fingerprint table
- [ ] Returns explicit `H0 / UNKNOWN` when nothing matches
- [ ] Prints the delta between last app-log line and cancel time (co-incidence test)
- [ ] 5 unit tests pass: forced-stop→Trigger 2, phantom-touch→H3, lid-leak→H1, no-fingerprint→H0, timestamp delta (`pytest unit_tests/test_scan_cancel_logs.py -v`)
- [ ] Runs against native (`logs/`) or Docker (`data/logs/`) paths via args

## Blocked by

None - can start immediately.

PR #168, opened 2026-06-16
Implement snapshot analysis

### LINKED ISSUE #160: Build snapshot_resources.py — per-run resource snapshot + diff (Trigger 2 disambiguator)
## What to build

A stdlib-only script that captures a labeled snapshot of host/process metrics once per run and, in `--report` mode, prints the run-over-run **diff table** — the metric that climbs monotonically outs the cause. Per `specs/analysis/sentri-self-cancel-study.md` §6.3. Degrades gracefully when a source is absent.

Captures: controller thread count, controller+app open FDs, TIME_WAIT sockets, controller+app RSS, SoC temp + throttled flags (`vcgencmd`, host only), disk free, `history.json` size, `/button_status` latency p95, web-app container uptime, recent OOM lines.

## Triggers addressed

Disambiguates **within Trigger 2 (safety net)** — answers *why* the web app went unreachable:
- **H4 (SoC thermal throttle)** — temp climbing / `get_throttled` non-zero
- **H5 (FD/socket leak)** — open-FD or TIME_WAIT count climbing
- **H6 (memory/OOM)** — RSS climbing / OOM in dmesg
- **H2 (event-loop block)** — `/button_status` p95 spiking, `history.json` growing
- Also surfaces **H1** (controller thread count) and **H7** (container uptime reset)

Does **not** speak to Trigger 1 — that's scan_cancel_logs' job.

## Acceptance criteria

- [ ] Appends one labeled row per invocation to a CSV
- [ ] Captures every metric in the §6.3 table; missing sources degrade gracefully (no crash)
- [ ] `--report` prints a column-aligned run-over-run diff
- [ ] Flags any metric that increases on every successive run as a prime suspect
- [ ] Treats `/button_status` timeouts as the latency ceiling (not a crash)
- [ ] Runs on the Pi host (for `vcgencmd`) with native or Docker PIDs via args

## Blocked by

None - can start immediately.

jackhu23:
## What Changed

Snapshot_resources.py: a script that captures a labeled snapshot of host/process metrics (thread count, etc.) once per run and prints a run-over-run difference table. Helps us identify why a safety net is fired if Trigger 2 is the culprit. 

## Why

https://github.com/AcornGenetics/aquilla-main/issues/160

## Spec

`specs/analysis/snapshot-resources.md`

## Changes Made

- [x] Source code modified
- [x] New tests written

## Hardware Testing

## Reviewer Notes

### DIFF
--- a/scripts/diagnostics/snapshot_resources.py
+++ b/scripts/diagnostics/snapshot_resources.py
@@ -187,6 +187,20 @@ def _button_p95(app_url, n=20):
     return "%.3f" % lat[int(0.95 * len(lat)) - 1]
 
 
+# Where the run history file may live, most-specific first. The real device
+# path is logs/history.json (aquila_web.main: BASE_DIR/"logs"/"history.json");
+# data/ and bare are fallbacks for Docker/other layouts.
+HISTORY_CANDIDATES = ("logs/history.json", "data/history.json", "history.json")
+
+
+def first_existing_size(candidates):
+    """Size (as str) of the first existing path in candidates, else ''."""
+    for cand in candidates:
+        if os.path.exists(cand):
+            return str(os.path.getsize(cand))
+    return ""
+
+
 def collect(label, backend_pid=None, app_pid=None, app_url=None):
     """Gather one snapshot from the host. Missing sources yield blank cells."""
     statvfs_free = ""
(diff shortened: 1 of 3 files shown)


NicoleCornell8 (dismissed):
## One bug to fix, then good to merge

### Bug: `history_bytes` will always be blank on a real device

`collect()` searches these candidates:
```python
for cand in ("data/history.json", "history.json"):
```

But the actual path on a Sentri is `logs/history.json` — set in `aquila_web/main.py` as `BASE_DIR / "logs" / "history.json"`. Neither candidate matches, so the column will always be empty in the field. Since this metric specifically targets H2 (history.json growing large enough to cause blocking I/O), a permanently blank cell defeats its purpose.

Fix:
```python
for cand in ("logs/history.json", "data/history.json", "history.json"):
```

---

### Minor notes (no block)

**`_button_p95` worst-case runtime:** 20 requests × 6s timeout + 0.1s sleep = up to ~2 minutes. Worth a one-line comment so an operator running this during an active investigation knows to expect the wait.

**Everything else looks good.** Clean separation of testable logic from on-device I/O glue, consistent with #158 and #159. Test coverage is solid — `monotonic_trend`, CSV round-trip, `analyze` flags and non-flags, blank/non-numeric graceful degradation all covered. `config.json` / `config.py` not touched.

PR #182, opened 2026-06-19
Implement new heirarchy change

### LINKED ISSUE #181: Well Verdict: Detected-wins precedence + dual-channel pill label on History detail
## What to build

On the **History detail** page, change how a Well's two Channel Calls aggregate into a single result. Two behaviour changes, both in `aquila_web/static/history_detail.js`:

1. **Well Verdict precedence → `Detected > Inconclusive > Not Detected`** (reversed from today's Inconclusive-wins). A Well is Detected if any Channel is Detected; else Inconclusive if any is Inconclusive; else Not Detected. The verdict sets the pill colour.
2. **Pill text names both Channels.** Group channels by Call in precedence order, FAM before ROX within a group, joined with ` + `; groups separated by a space; channels named by the profile dye label (FAM/ROX fallback); negative string is `Not Detected`. E.g. `Detected (FAM) Inconclusive (ROX)`, `Not Detected (FAM + ROX)`.

KPIs count by verdict, mutually exclusive (a `Detected+Inconclusive` Well → Detected only). **QC Status stays channel-sensitive**: `Review` if *any* Channel is Inconclusive (excluding `ROX Unavailable`), independent of the verdict — so a Detected-overall Well with an inconclusive Channel still flags Review. `ROX Unavailable` is excluded from both verdict and text (FAM shown alone).

Scope: History detail page only — the live run page (`script.js`) is unchanged. Full detail in `specs/analysis/spec_well_verdict_precedence.md`; rationale in `docs/adr/ADR-015-well-verdict-detected-wins-precedence.md`.

## Acceptance criteria

- [ ] Well Verdict resolves by `Detected > Inconclusive > Not Detected`; pill colour matches the verdict
- [ ] Pill text names both channels, grouped/ordered per spec, using profile dye labels with FAM/ROX fallback and `Not Detected` wording
- [ ] `Detected(FAM)+Inconclusive(ROX)` → pill `Detected (FAM) Inconclusive (ROX)`, Detected KPI +1 (not Inconclusive), **QC Status: Review**
- [ ] `ROX Unavailable` excluded from verdict and text; FAM shown alone (e.g. `Detected (Target)`)
- [ ] Detected/Inconclusive KPI buckets are mutually exclusive (sum <= 4)
- [ ] Unit tests (node:test) cover the verdict/label/KPI/QC cases against `summarizeResults`
- [ ] Live run page (`script.js`) behaviour unchanged

## Blocked by

None - can start immediately

jackhu23:
## What Changed

Changed the hierarchy of detection [Old: Inconclusive > Detected > Undetected] --> [New: Detected > Inconclusive > Undetected]. Furthermore, changed the Tube UI in run history to showcase both FAM & ROX regardless. EX. FAM detected + ROX detected now shows Detected (FAM) Detected (ROX) rather than just detected. Or FAM Detected ROX Inconclusive shows Detected (FAM) Inconclusive (ROX). 

## Why

https://github.com/AcornGenetics/aquilla-main/issues/181

## Spec

`specs/analysis/spec_well_verdict_presedence.md`

## Changes Made

- [x] Source code modified
- [x] New tests written
- [x] ADR written (if an irreversible architectural decision was made)

## Hardware Testing

- [x] Tested in simulation mode only — reason: [why physical not possible]

## Reviewer Notes

Previously the hierarchy was Inconclusive > Detected, so any inconclusive channel made the whole well's verdict Inconclusive — which meant QC Status was already effectively channel-sensitive (it read Review whenever any channel was inconclusive, and Pass only when every well was Detected or Not Detected).

With the hierarchy now Detected > Inconclusive, that's no longer automatic. A well that's Detected on one channel and Inconclusive on the other now resolves to an overall verdict of Detected. To avoid letting a Detected verdict mask an inconclusive channel, I deliberately kept QC Status channel-sensitive: it still reads Review if any channel is inconclusive, even when the overall verdict is Detected.

If you'd prefer QC to follow the overall verdict instead (i.e. Pass when the verdict is Detected, regardless of an inconclusive channel), let me know and I'll change it.


nana-acorn:
Reviewed the logic and ran the suites. **The change looks correct — no blocking issues.**

**Testing (verified locally):**
- Trial-merged `main` into this branch — clean, no conflicts. Integrated suite: **558 passed / 6 failed / 134 skipped**. The 6 failures are pre-existing on `main` (`test_wifi_helpers.py` nmcli mocks + `test_linux_compatibility`) and unrelated to this PR — **no new failures introduced**.
- The verdict logic itself is fully green (20 JS tests). I added a set of tripwire tests for the precedence/QC/divergence behaviours.
- The new precedence flips **exactly** the two mixed Detected+Inconclusive cells — nothing else changes. The QC channel-sensitivity is implemented correctly: I swept all 16 FAM×ROX combos and QC reads Review whenever any scored channel is Inconclusive, so a Detected verdict never silently masks an inconclusive channel. 👍 on keeping QC channel-sensitive — agreed that's the safe call for a diagnostic device.

**Minor / non-blocking:**
- **ADR contradicts some dead code in `aquila_web/static/script.js` (not code this PR touches).** ADR-017 (lines 39–40) says the live run page "shows two independently coloured half-dots and a per-Channel table, **with no aggregated verdict**." That's true of what's *rendered*, but `script.js` (lines 928–949) still computes a per-tube `tubeDetected`/`tubeInconclusive` aggregation using the old inconclusive-wins precedence — those arrays are just never read by the render below them. So the ADR statement contradicts the (dead) code. Not a real concern since nothing displays it; the clean fix is to just delete that dead block, after which the ADR is accurate as-is. Fine as a follow-up.
- Spec vs. implementation mismatch on tests: the spec says `summarizeResults` isn't exported and prescribes Playwright e2e (citing ADR-003), but the implementation adds a guarded CommonJS export + `node:test` unit tests instead. The export is browser-safe (verified it loads under a no-`module` sandbox without throwing or leaking globals), so I'm fine with the approach — just update the spec's Tests section to match what was built, and note the ADR-003 carve-out.
- These JS unit tests aren't wired into CI (no `node --test` in `.github/`), so they're manual-only and can rot. Worth a follow-up to add them to CI.
- Branch is a few commits behind `main` — rebase/merge before landing (it's clean).

NicoleCornell8 (approved):
Approved — logic is correct and the tests pass (verdict precedence, dual-channel labels, ROX Unavailable exclusion, and channel-sensitive QC all check out).

One note: the spec/ADR describe a Playwright e2e testing approach, but the implementation went with `node:test` unit tests instead (which is honestly the cleaner path for this pure logic). Worth reconciling the spec's Tests section so the docs match what shipped. Not a blocker.

NicoleCornell8 (commented):
Re-reviewed after the latest commits. The verdict-precedence logic is unchanged from the earlier approval and the unit tests still pass (9/9). The new commits since then are housekeeping — version bump, main merge, help.html rollback — no change to the feature, so the approval still stands on the code.

One new thing to sort out before merge: the ADR number collides. This PR adds `docs/adr/ADR-017-well-verdict-detected-wins-precedence.md`, and PR #194 (optics-flash fix) also adds an `ADR-017` (`ADR-017-async-gated-ui-defaults-to-hidden.md`). Both are open against main, so whichever lands second will clash. I've already asked #194 to take the next free number — just flagging it here too so you two coordinate and don't both keep 017. The commit that renamed this from ADR-015 to ADR-017 is what introduced the overlap.

Nothing else blocking from me.

NicoleCornell8:
merge

PR #184, opened 2026-06-19
Implemented auto reboot on device update

### LINKED ISSUE #183: OTA: auto-reboot after update completes + one-time Update Complete modal
## Parent

#120 (Fix updates complete page) — this is the chosen-approach implementation ticket for that bug.

## What to build

After a successful manual OTA update, the device should **reboot itself** (full software reboot) and, once it comes back up, show a **one-time, blocking "✓ Update Complete" modal** that the operator dismisses with **OK**. This removes the indefinite frozen-update-screen and the manual power-cycle that recovery currently requires.

End-to-end behaviour:

1. `/update/apply` writes a two-state sentinel to `/opt/fleet/last_update.json` (`reboot_pending`), then triggers Watchtower as today (the run-active 409 guard stays).
2. Watchtower pulls the new image and swaps the containers. No reboot happens here (pull in progress; the applying container is killed in the swap).
3. The **new container**, on startup, sees `reboot_pending` → flips the sentinel to `show_complete` (persisted before rebooting) → calls a new `kiosk-control /reboot` endpoint → host runs `systemctl reboot`.
4. The Pi reboots, Chromium relaunches to the login screen.
5. After the operator logs in, the first post-login page (via `nav.js` → `/update/status` returning `complete`) shows a blocking "✓ Update Complete" modal.
6. Tapping **OK** calls `/update/ack-complete`, which deletes the sentinel and resets status. The modal shows exactly once — no boot loop, no repeat.

Full design, edge cases, and rationale: `specs/backend/spec_ota_auto_reboot_complete.md` and ADR-016. Supersedes `specs/frontend/spec_ota_update_complete_on_reload.md`.

**Rollout note:** `kiosk_control.py` is a host systemd service, not part of the Docker image Watchtower updates. The new `/reboot` endpoint reaches devices only via the host-service deploy path (`update_host_service.sh` / `deployment2.sh`), so hosts must be updated before auto-reboot works. If the reboot call fails, the sentinel stays `show_complete` so the modal still shows on the next manual reset (degrades to no-worse-than-today).

## Acceptance criteria

- [ ] A successful OTA update triggers exactly one full device reboot (no boot loop)
- [ ] After reboot + login, a blocking "✓ Update Complete" modal appears on the first post-login screen and is dismissed with OK
- [ ] The modal shows exactly once; it does not reappear on subsequent normal boots
- [ ] `/update/status` exposes a `complete` state; `POST /update/ack-complete` clears the sentinel and resets status
- [ ] New `POST /reboot` endpoint on kiosk-control runs `systemctl reboot`, origin-restricted to loopback/Docker bridge
- [ ] The reboot is skipped if a run is active (defensive guard)
- [ ] If kiosk-control is unreachable, the update still records completion (modal shows on next reset) and the failure is logged — never worse than current behaviour
- [ ] Update failures still show the existing error state; no reboot on failure
- [ ] Tests pass: pytest tests unit_tests -v (sentinel state machine unit tests; /update/status + ack-complete contract tests; reboot subprocess mocked / marked hardware)

## Blocked by

None - can start immediately

jackhu23:
## What Changed

Implemented a SENTRI auto reboot on device update. When the user updates their machine, once the new image has been successfully downloaded and the old image tossed, the device will automatically reboot. On device startup after the reboot, the user will be prompted with a flag detailing that their device has been updated.

## Why

[Link to the GitHub issue: Closes #[number]](https://github.com/AcornGenetics/aquilla-main/issues/183)

## Spec

`specs/frontend/spec_ota_update_complete_on_reload.md`

## Changes Made

- [x] Source code modified
- [x] New tests written
- [x] ADR written (if an irreversible architectural decision was made)

## Hardware Testing

## Reviewer Notes

MUST BE TESTED ON A PHYSICAL DEVICE. I.E. FIX HAS NOT BEEN TESTED YET

### DIFF
--- a/aquila_web/main.py
+++ b/aquila_web/main.py
@@ -1518,6 +1518,58 @@ _startup_image_digest_ui: str | None = os.getenv("RUNNING_IMAGE_DIGEST_UI") or N
 _latest_ghcr_digest: str | None = None
 _latest_ghcr_digest_ui: str | None = None
 
+# --- OTA auto-reboot completion sentinel (issue #183, ADR-016) ----------------
+from aquila_web import update_sentinel as _sentinel
+
+# Lives on the /opt/fleet host volume so it survives the container swap + reboot.
+_UPDATE_SENTINEL_PATH = os.getenv("AQ_UPDATE_SENTINEL_PATH", "/opt/fleet/last_update.json")
+# Short TTL: a sentinel older than this is stale and must not pop a modal.
+_UPDATE_SENTINEL_TTL = int(os.getenv("AQ_UPDATE_SENTINEL_TTL", "600"))
+
+
+def _utcnow_iso() -> str:
+    return datetime.utcnow().isoformat() + "Z"
+
+
+def _trigger_host_reboot() -> bool:
+    """Best-effort: ask the host kiosk-control service to reboot the device.
+
+    Synchronous + swallows errors: the process is about to die anyway, and a
+    failure must not crash startup (it degrades to 'modal on next manual reset').
+    """
+    try:
+        headers = {"Authorization": f"Bearer {WATCHTOWER_TOKEN}"} if WATCHTOWER_TOKEN else {}
+        with httpx.Client(timeout=10.0) as client:
+            client.post(f"{KIOSK_CONTROL_URL}/reboot", json={}, headers=headers)
+        return True
+    except Exception as e:  # noqa: BLE001 - never let a reboot failure crash boot
+        logger.warning("host reboot request failed: %s", e)
+        return False
+
+
+def _resolve_startup_update_state() -> None:
+    """On boot, advance the update sentinel state machine.
+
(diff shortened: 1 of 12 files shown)


NicoleCornell8 (commented):
Tests pass (13/13), boot-loop safety verified, kiosk-control /reboot is origin-checked. A few notes:

1. (Main one) `/update/apply` writes the `reboot_pending` sentinel before calling Watchtower but doesn't clear it on failure (non-200 or exception). After a *failed* update, if the container restarts within the TTL it'll reboot + show 'Update Complete'. Clear the sentinel in both failure branches.
2. `POST /reboot` proxy is unauthenticated and unused by the frontend — drop it or note why kept.
3. `datetime.utcnow()` is deprecated (3.12+); update_sentinel.py is tz-aware but main.py isn't.

Also: you flagged it's untested on hardware — the real `systemctl reboot` and the two-stage host-service rollout are the risky bits, don't fleet-deploy on green CI alone.

PR #189, opened 2026-06-19
Rebrand #186 + #187: full aquila→sentri code + deployment-pipeline rename

nana-acorn:
Closes #186 and #187. Part of the aquila → sentri rebrand (#178, PRD #185, ADR-015). The HITL fleet cutover is #188 (not in this PR).

## #186 — internal code rename
`aq_curve`/`aq_lib`→`sentri_*`, `aquila_web`→`sentri_web` (incl. dynamic `importlib`), brand **word** `Aquila`→`Sentri` (case-preserving). Guarded by `tests/unit/test_rebrand_completeness.py` (package tokens + word-boundary brand check).

## #187 — deployment pipeline (TDD per dimension, `tests/fleet_device/`)
- **D1** container_name → `sentri-*` + all service-DNS refs in lockstep (`test_compose_dns_integrity`)
- **D2** systemd units renamed + every deploy/verify-script ref updated (`test_service_unit_integrity`)
- **D3** image defaults → `acorngenetics/sentri`; CI derives from `$GITHUB_REPOSITORY` (`test_image_path_agreement`)
- **D4** selective-rename guard: `/opt/aquila` + `AQ_` carve-outs preserved (`test_selective_rename_guard`)
- **D5** `STATE_DIR` parameterization for cheap future host migration (`test_state_dir_param`)
- Swept all deployment-identifier strings across scripts/docs/tests; existing fleet_device tests updated to new names
- **CI dual-publish** to old `aquilla-main-*` path — TEMPORARY, removed in #188

**Carve-outs kept** (ADR-015): `AQ_` env vars, `/opt/aquila`.

## Verification
- Completeness guard + D1–D5: green. Suite: **563 passed**; 7 failures **pre-existing** (wifi/lid-worker/`/dev`-mount compose test), identical on a clean tree — behavior-preserving.
- Branch Docker build (api+ui) dispatched and green (#186); re-dispatched after #187.

## CI caveats (this repo)
- No `pull_request` CI / no pytest/e2e workflow — this PR runs nothing automatically; guard/suite run locally. (Follow-up: add a `pytest` workflow.)
- e2e/Playwright not covered by any CI.

🤖 Generated with [claude-flow](https://github.com/ruvnet/claude-flow)

PR #194, opened 2026-06-23
Fix optics path field flashing on Run page in non-dev mode (#193)

### LINKED ISSUE #193: Optics path field flashes on Run page in non-dev mode (FOUC)
## Summary

In non-dev mode, the dev-only Optics path field flashes visible on the Run page for a moment on every navigation to `/run` before JavaScript hides it.

## Environment

- **Device type:** Mac simulation
- **Simulation mode:** No — bug occurs in **non-dev** mode (`AQ_DEV_SIMULATE=0`), which is what end users see
- **Branch/commit:** `sentri-rebrand` @ `0fdce8c`
- **Python version:** 3.13.2

## Steps to Reproduce

1. Run the server in non-dev mode (user-facing config):
   `AQ_DEV_SIMULATE=0 AQ_DEV_RUN_DURATION=3 uvicorn sentri_web.main:app --host 127.0.0.1 --port 8090`
2. Navigate to another page, then click back to the Run page (`/run`).
3. The "Optics:" path field appears briefly in the run header, then disappears.

## Expected Behavior

The Optics path field is dev-only. In non-dev mode it should **never** be visible — not even for a frame. (In dev mode it is always shown on the Run page, which is correct.)

## Actual Behavior

The field flashes visible on first paint, then vanishes once JS resolves the dev state. It goes away fast but should not appear at all.

## Root Cause

`run.html` is served as a **static file** (no server-side templating), so the field's visibility is decided client-side. The markup shipped `#run-optics-tab` as `class="run-optics-tab"` — **visible by default**. The only thing that hides it is the callback of an **async** `fetch("/button_status")`, which runs `setOpticsVisibility(false)` in non-dev mode. Between first paint and the fetch resolving, the field is visible → the flash. Each navigation reloads `run.html` fresh, so it flashes every time.

This is a flash-of-unhidden-content (FOUC): visibility gated on async state, but defaulting to *shown*.

## Fix

Ship the field hidden by default in static markup; JS reveals it only in dev mode (`setOpticsVisibility` uses a forced `classList.toggle("is-hidden", !isDev)`, so it reveals in dev and is a no-op in non-dev):

```html
<div class="run-optics-tab is-hidden" id="run-optics-tab">
```

- File changed: `sentri_web/static/run.html`
- Regression test added: `tests/contract/test_optics_path_endpoints.py::test_run_page_hides_optics_tab_by_default` — GETs `/run` and asserts the optics tab ships with `is-hidden`. Red before fix, green after.

## Follow-up

The underlying convention is being codified in **ADR-017**: UI elements whose visibility depends on async state must default to hidden in markup and be revealed by JS — never the reverse.

nana-acorn:
Closes #193.

## Problem
The dev-only Optics path field (`#run-optics-tab`) shipped **visible** in the static `run.html` and was only hidden after the async `fetch("/button_status")` callback resolved. In non-dev mode (`AQ_DEV_SIMULATE=0` — what users see) this caused a flash-of-unhidden-content on every navigation to `/run`.

## Fix
Ship the field hidden by default (`is-hidden`). `setOpticsVisibility(isDev)` uses a forced `classList.toggle("is-hidden", !isDev)`, so it reveals the field in dev mode and is a no-op in non-dev — no flash, and a fail-safe default (gated UI stays hidden if the fetch fails).

## Changes
- `aquila_web/static/run.html` — add `is-hidden` to `#run-optics-tab`
- `tests/contract/test_optics_path_endpoints.py` — regression test asserting the served `/run` markup hides the optics tab by default (red before fix, green after)
- `docs/adr/ADR-017-async-gated-ui-defaults-to-hidden.md` — records the convention: async-gated UI defaults to hidden in markup

## Verification
- Reproduced in non-dev mode; served HTML now ships `is-hidden`
- Dev mode still shows the field (forced toggle)
- `tests/contract` + `tests/e2e/test_run_dropdowns.py` + `unit_tests/test_optics_history.py`: 120 passed, 3 skipped

### DIFF
--- a/docs/adr/ADR-017-async-gated-ui-defaults-to-hidden.md
+++ b/docs/adr/ADR-017-async-gated-ui-defaults-to-hidden.md
@@ -1,0 +1,75 @@
+# ADR-017: Async-gated UI elements default to hidden in markup
+
+**Status:** Accepted
+**Date:** 2026-06-23
+**Author:** Rasita Vajapattana
+**Deciders:** Rasita Vajapattana
+
+> Numbering note: this branch is cut from `main` (latest ADR-015). ADR-016 is
+> reserved by the in-flight rebrand branch (#189), so this decision takes 017
+> to avoid a collision regardless of merge order.
+
+---
+
+## Context
+
+The web UI (`aquila_web`) serves its pages as **static HTML files** via FastAPI `FileResponse` — there is no server-side templating. Any element whose presence depends on runtime state (dev vs. non-dev mode, feature flags, fetched config) therefore has its visibility decided **client-side**, after an asynchronous fetch resolves.
+
+The Run page has a dev-only "Optics path" field (`#run-optics-tab`). It was authored visible in the static markup, and JS hid it after `fetch("/button_status")` resolved with `dev_simulate=false`. Because the fetch is async, the field painted visible on first frame and was only hidden once the callback ran — a flash-of-unhidden-content (FOUC) that end users saw on every navigation to `/run` in non-dev mode (see issue #193).
+
+The constraints that make this a recurring trap:
+
+- Static markup paints **before** any JS runs; the initial class list is what the user sees first.
+- Fetch-gated visibility means there is always a gap between first paint and the state-resolving callback.
+- Defaulting to *shown* means the failure mode leaks privileged/dev-only UI to users; defaulting to *hidden* means the worst case is a correct-but-slightly-late reveal.
+
+Doing nothing leaves every current and future async-gated control exposed to the same flash.
+
+## Decision
+
+**We will author every UI element whose visibility depends on async/runtime state as hidden by default in static markup (`is-hidden`), and reveal it from JS only once the state is known.**
+
+Concretely:
+
+- New dev-only, flag-gated, or fetch-gated elements ship with `class="... is-hidden"` in the `.html` file.
+- The JS that resolves the state uses a **forced** toggle, e.g. `el.classList.toggle("is-hidden", !shouldShow)`, so it reveals when the condition holds and is a harmless no-op when it does not.
+- The reference fix: `#run-optics-tab` in `aquila_web/static/run.html` now ships `is-hidden`; `setOpticsVisibility(isDev)` reveals it only in dev mode.
+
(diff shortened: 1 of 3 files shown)


nana-acorn:
## Test verification: no new bugs introduced ✅

Ran the full suite (`tests` + `unit_tests`) on this branch vs `main` from **identical fresh worktree checkouts** for an apples-to-apples comparison:

| | Fresh `main` (baseline) | This branch |
|---|---|---|
| Failed | 34 | 34 |
| Errors | 64 | 64 |
| Passed | 531 | **532** |
| Skipped | 134 | 134 |

- **FAILED sets: identical.** **ERROR sets: identical.** Every failing/erroring test on this branch already fails on `main`, untouched by this change.
- The only delta is **+1 passing test** — the new regression test added here: `tests/contract/test_optics_path_endpoints.py::test_run_page_hides_optics_tab_by_default`, which GETs `/run` and asserts the optics tab ships with `is-hidden` (red before the fix, green after).

The pre-existing red is all environmental and unrelated to the optics flash: `tests/e2e/*` (Playwright — needs a live browser/frontend), missing result/curve fixture files (`FileNotFoundError`), hardware fakes (`_FakeSMBus`), and exhausted wifi mocks (`StopIteration`).

NicoleCornell8 (approved):
Approved. Correct fix — ships `is-hidden`, `setOpticsVisibility` reveals via forced toggle, fail-safe default. Test passes.

One thing: ADR number collision — PR #182 (heirarchy-check) also adds an `ADR-017`. Your numbering note covered #189's 016 but missed #182's 017. Bump this to ADR-019 (or next free) so they don't clash on merge.

docs/adr/ADR-017-async-gated-ui-defaults-to-hidden.md
```diff
@@ -0,0 +1,75 @@
+# ADR-017: Async-gated UI elements default to hidden in markup
```
NicoleCornell8, line 1: ADR number collision: PR #182 (`heirarchy-check`) also adds an `ADR-017` (`ADR-017-well-verdict-detected-wins-precedence.md`). The numbering note above accounts for #189's reserved 016 but misses #182's 017 — they'll clash whichever merges second. Bump this to the next free number (019) and rename the file.

Issue #201, opened 2026-06-25
Structured Profiles A3 — wire `stages` into POST/GET profile endpoints (contract)


jackhu23:
**Route:** backend (Dev A) · **Seam:** contract (`tests/contract/`) · **Type:** AFK
**Spec:** `specs/backend/spec_profile_endpoints_wiring.md` · **PRD:** `specs/prd/structured-profile-editor-prd.md` · **Decision:** ADR-018

## What to build

Wire assembly (A1) and validation (A2) into the HTTP surface so a structured Profile round-trips through `POST`/`GET`.

- `POST /profiles` with a `stages` payload → run validation; on success, write a Profile whose JSON carries both the `stages` object (source of truth) and the regenerated `steps`; on failure, return the existing 4xx error convention and write nothing.
- `GET /profiles` returns a `structured` boolean per Profile (true iff its JSON has `stages`) so the list can route each row.
- `GET /profiles/details` returns the `stages` object when present so the editor can repopulate.
- `steps` remains the only artifact the runner/analytics read; `stages` is additive.

## Acceptance criteria

- [ ] POST with a valid `stages` payload → 200; written JSON has correct `steps` and a round-trippable `stages`.
- [ ] POST with out-of-range / blank / wrong-sub-stage-count `stages` → error, nothing written.
- [ ] `GET /profiles` returns `structured: true` for a structured Profile and `false` for a Legacy Profile.
- [ ] `GET /profiles/details` returns `stages` for a structured Profile and omits it for a Legacy Profile.
- [ ] Contract tests added to `tests/contract/test_profile_endpoints.py` (prior art there + `test_bundled_profiles.py`).
- [ ] Bundled Profiles remain read-only (existing 403 behavior unchanged).

## Blocked by

- #198
- #199

jackhu23:
**Carry-over from A1 (#198 / PR #206) review:**

`assemble_steps` assumes well-formed input — it reads `stages['incubation']['enabled']`, `stages['amplification']['subStages']`, etc. directly, so malformed input raises `KeyError` rather than a clean error. **A3 must run `validate_stages` (A2 #199) and return the 4xx BEFORE calling `assemble_steps`**, so the assembler is only ever invoked on validated `stages`.

jackhu23:
🔗 Implemented by PR #209 — merged into `updated-profiles`. Will auto-close when `updated-profiles` is promoted to `main` via the #204 integration PR.

PR #208, opened 2026-06-25
#199 (A2) validate_stages

### LINKED ISSUE #199: Structured Profiles A2 — validate `stages` (ranges + sub-stage count)
**Route:** backend (Dev A) · **Seam:** unit (`unit_tests/`) · **Type:** AFK
**Spec:** `specs/backend/spec_stages_validation.md` · **PRD:** `specs/prd/structured-profile-editor-prd.md` · **Decision:** ADR-018

## What to build

A pure validation function over a `stages` object that returns the set of validation errors (per field), used by the backend to reject invalid saves. No HTTP — pure input → errors.

Rules:
- Temperature 25–100 °C (every editable temp field).
- Time 1–600 s (every editable time field).
- The Extension-bearing Sub-stage time is 11–600 s (so the optics split stays valid).
- Cycles: integer 1–50.
- Amplification Sub-stage count is 2 or 3.
- Disabled (unchecked) Stages are skipped — their values never produce errors.
- Non-numeric / blank values for enabled fields are errors.

## Acceptance criteria

- [ ] Pure validation function flags out-of-range temp, time, extension-time, and cycles.
- [ ] Disabled Stages are skipped entirely.
- [ ] Sub-stage count outside 2–3 is an error.
- [ ] Blank / non-numeric enabled fields are errors.
- [ ] Unit tests in `unit_tests/` cover each boundary (25/100 °C, 1/600 s, extension 11 s, cycles 1/50, sub-stage count).

## Blocked by

- #197

jackhu23:
Closes #199 — A2: backend range/shape validation for structured `stages`.

## What this does
Adds `validate_stages(stages) -> list[str]` to `aquila_web/profile_assembly.py` — the backend's defense-in-depth check. Returns a list of field-named error strings (empty = valid); does not reject the save itself (that's A3 #201).

Rules (enabled stages only; disabled optional stages skipped):
- temp `25–100 °C`; time `1–600 s`
- **extension-bearing sub-stage time `≥ 11 s`** (`OPTICS_TAIL_SECONDS + 1`) so the `assemble_steps` `(time-10)` split never goes negative/zero — closes A1-review note #1
- cycles integer `1–50`; sub-stage count `2–3`
- non-numeric / blank values are errors, not exceptions

Also carries the A1-review cosmetic fix: `assemble_steps` annotated `-> list[dict]` (note #3).

## Spec / docs
- Spec: `specs/backend/spec_stages_validation.md` · PRD: `specs/prd/structured-profile-editor-prd.md` · ADR-018

## SPEC_WORKFLOW checklist
- [x] Spec linked + accurate
- [x] New tests (8 validation tests, RED→GREEN)
- [x] `pytest unit_tests/test_profile_assembly.py -v` → 19 passed
- [x] No regressions — only `profile_assembly.py` + its test touched; `main.py` untouched. The 3 pre-existing contract failures are environmental and unchanged from baseline.
- [x] No secrets

## Out of scope
Enforcing on `POST /profiles` (call `validate_stages` before `assemble_steps`, return 4xx) — A3 (#201), tracked by the carry-over comment there.

### DIFF
--- a/aquila_web/profile_assembly.py
+++ b/aquila_web/profile_assembly.py
@@ -9,8 +9,61 @@ unit-testable in isolation. See specs/backend/spec_profile_step_assembly.md
 # extension-bearing sub-stage is split into (time - tail) -> optics -> tail (ADR-018).
 OPTICS_TAIL_SECONDS = 10
 
+TEMP_MIN, TEMP_MAX = 25, 100
+TIME_MIN, TIME_MAX = 1, 600
+# The extension-bearing sub-stage holds for (time - OPTICS_TAIL_SECONDS) before the
+# optics read, so its time must leave at least 1 s — hence a higher floor.
+EXTENSION_TIME_MIN = OPTICS_TAIL_SECONDS + 1
+CYCLES_MIN, CYCLES_MAX = 1, 50
 
-def assemble_steps(stages: dict) -> list:
+
+def _is_number(value) -> bool:
+    """True for a real numeric value (rejects bools, strings, None)."""
+    return isinstance(value, (int, float)) and not isinstance(value, bool)
+
+
+def validate_stages(stages: dict) -> list[str]:
+    """Check a structured `stages` object against the instrument's valid ranges.
+
+    Returns a list of human-readable error strings (each naming the offending
+    field); an empty list means valid. Disabled optional Stages are skipped.
+    Pure check only — enforcement on POST is A3 (#201).
+    """
+    errors = []
+
+    def check_temp(value, field):
+        if not (_is_number(value) and TEMP_MIN <= value <= TEMP_MAX):
+            errors.append(f"{field}.temp: Invalid Value")
+
+    def check_time(value, field, minimum=TIME_MIN):
+        if not (_is_number(value) and minimum <= value <= TIME_MAX):
+            errors.append(f"{field}.time: Invalid Value")
+
+    for key in ("incubation", "denaturation", "finalHold"):
+        stage = stages[key]
+        if not stage.get("enabled"):
+            continue
+        check_temp(stage.get("temp"), key)
+        check_time(stage.get("time"), key)
+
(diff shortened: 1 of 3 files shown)


jackhu23:
Addressed both findings in 643c390 — `validate_stages` is now treated as the trust boundary its docstring promises (never raises on a dict input):

- **Finding #1 (subStages crash):** `subStages` that is `null` / non-list / wrong length now appends `amplification.subStages: Invalid Value` and returns early; non-dict sub-stage elements are flagged per-index and skipped instead of `.get()`-ing on a non-dict. No more TypeError/AttributeError.
- **Finding #2 (KeyError on missing keys):** optional stages use `.get()` + `isinstance(dict)` (missing/malformed → skipped, can't be 'enabled'); a missing/non-dict `amplification` returns `amplification: Invalid Value` instead of KeyError.
- **Minor note:** added `_is_int` and reused it for the cycles integer check.

Added 4 regression tests (`test_malformed_substages_*`, `test_substages_with_non_dict_elements_*`, `test_missing_amplification_*`, `test_missing_optional_stage_key_*`). Suite now 23 passing. Ready for re-review.

jackhu23:
Closed the residual from finding #1 in the latest commit — `validate_stages` now requires every key `assemble_steps` dereferences, so a validate-clean payload is guaranteed safe to assemble:

- **missing sub-stage `name`** → `amplification.subStages[i].name: Invalid Value` (was: validate `[]` → `KeyError: 'name'`)
- **optional stage missing `enabled` / present-but-non-dict / missing key** → `<stage>: Invalid Value` (was: silently skipped → `KeyError`/`TypeError` in assembly). `enabled` must now be present and boolean even on disabled stages, since `assemble_steps` subscripts it.

Added 5 regression tests, including `test_validate_clean_implies_assemble_does_not_raise` which asserts the trust-boundary guarantee directly. Suite now 27 passing. The three probe shapes you listed all return errors now (verified). Ready for re-review.

PR #217, opened 2026-06-26
Update motor_class.py

jake-acorn:
speed the drawer motor up!!!

## What Changed

Changed the timing on motor_class.py to make drawer a bit faster on open and close
## Why

Link to the GitHub issue: Closes #[number]

## Spec

`specs/[type]/[filename].md` -- unsure what this means

just changing a motor_class attribute 

## Changes Made

- [ X] Source code modified
- [ ] Spec updated to match implementation
- [ ] New tests written
- [ ] Existing tests still pass
- [ ] No secrets or `.env` files in this diff

## Hardware Testing

## Reviewer Notes

### DIFF
--- a/aq_lib/motor_class.py
+++ b/aq_lib/motor_class.py
@@ -90,11 +90,11 @@ class Motor():
         delta = position - self.position
         return self.move_w_home_flag( delta, step_delay )
 
-    def move_abs_wo_home_flag( self, position, step_delay = 0.0 ):
+    def move_abs_wo_home_flag( self, position, step_delay = 0.0, pulse_delay = 0.0001 ):
         logger.info( "Moving %d", position )
         delta = position - self.position
         logger.info( "Moving delta %d", delta )
-        return self.move_wo_home_flag( delta, step_delay )
+        return self.move_wo_home_flag( delta, step_delay, pulse_delay )
 
     def set_dir( self, steps ):
         if steps < 0:
@@ -106,7 +106,7 @@ class Motor():
             self.direction = 0
             self.gpio.output( self.DIR_PIN, self.DIR_FORWARD_STATE)
 
-    def move_wo_home_flag( self, steps, step_delay = 0.0 ):
+    def move_wo_home_flag( self, steps, step_delay = 0.0, pulse_delay = 0.0001 ):
 
         self.set_dir( steps )
         self.enable()
@@ -115,7 +115,7 @@ class Motor():
             for k in range ( self.step_multiplier ):
                 self.gpio.output( self.STEP_PIN, HIGH)
                 self.gpio.output( self.STEP_PIN, LOW)
-                time.sleep ( 0.0001 )
+                time.sleep ( pulse_delay )
 
             time.sleep( step_delay )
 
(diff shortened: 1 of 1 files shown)


NicoleCornell8 (commented):
## Review summary

Two `step_delay` reductions on the Drawer (`open`: 0.0005→0.0003, `read`: 0.001→0.0005). The intent is a faster drawer, but I think this change won't accomplish much — and a couple of cleanups are needed.

### 🔴 The speedup is marginal
`step_delay` is only a small slice of the real per-step cost. In `move_wo_home_flag`, each logical step runs the inner micro-pulse loop `step_multiplier` times, each with a hard-coded `time.sleep(0.0001)`:

```
per-step ≈ step_multiplier × 100µs + step_delay
```

With `step_multiplier = 32` (host_config.json, sn01–03), the inner loop is ~3.2 ms/step, which dwarfs the delay being tuned. For `open_steps = 4500`:

| | per-step | total |
|---|---|---|
| Before (0.0005) | ~3.7 ms | ~16.6 s |
| After (0.0003) | ~3.5 ms | ~15.8 s |

≈ 5% (~0.9 s) faster on open, far less on read (only 152–250 steps). If the goal is a *noticeably* faster drawer, the lever is the inner `time.sleep(0.0001)` / how `step_multiplier` is applied — not `step_delay`.

### 🟡 'Close' isn't actually changed
The PR says faster on open **and close**, but the return/close path goes through `home()`, which uses a fixed `0.0020` delay this PR doesn't touch. So close speed is unchanged — either adjust `home()` too or update the description.

### 🟡 Hardware untested
Lower `step_delay` raises step frequency and can cause step-skipping under load (drawer not reaching full travel). Risk is bounded here since `open()`/`read()` re-`home()` each cycle and the delay is a small fraction of per-step time — but every checkbox is unchecked. Per repo rules, please bench on a physical unit (sn01–03), confirm full travel, and check the 'Tested on physical device' box before merge.

### ✅ Low-risk otherwise
Conservative positive values, no signature/API changes, trivially revertable, no secrets.

**Recommendation:** I'd hold for changes — fix the stale comments below, correct the 'close' claim, and verify on hardware. Happy to help retarget the inner loop if a real speedup is the goal.

aq_lib/motor_class.py
```diff
@@ -145,11 +145,11 @@ class Drawer ( Motor ):
 
     def open( self ):
         self.home()
-        ret = self.move_abs_wo_home_flag ( self.open_steps, 0.0005 ) #Changed to 0.0005 from 0.002 - Ryan 04/22/26
+        ret = self.move_abs_wo_home_flag ( self.open_steps, 0.0003 ) #Changed to 0.0005 from 0.002 - Ryan 04/22/26
```
NicoleCornell8: Stale comment: this still says `Changed to 0.0005 from 0.002 - Ryan 04/22/26` but the value is now `0.0003`. Please update it so the history stays accurate, e.g. `# 0.0003 — was 0.0005, Jake 06/26/26`.

aq_lib/motor_class.py
```diff
@@ -145,11 +145,11 @@ class Drawer ( Motor ):
...
 
     def read( self ):
         self.home()
-        ret = self.move_wo_home_flag ( self.read_steps, 0.001 )
+        ret = self.move_wo_home_flag ( self.read_steps, 0.0005 )
```
NicoleCornell8: This `read()` delay change (0.001 → 0.0005) has no comment. Add a short one matching the style on line 148. Note `read_steps` is only 152–250, so the time saved here is negligible (~0.1 s).

PR #227, opened 2026-06-26
#222: Images self-report their version (baked in + shown in footer)

### LINKED ISSUE #229: Images self-report their version (version baked in + shown in footer)
## Parent

#228

## What to build

A foundational tracer bullet: every built image carries a version and self-reports it end-to-end. The version is baked into the API image at build time and surfaced through the existing `/version` endpoint and the help footer — so a deployed Sentri shows what it is running, without anyone hand-editing the UI.

- The API image accepts a build-time version, exposes it as the `AQ_APP_VERSION` env var, and also records it as an `org.opencontainers.image.version` OCI label.
- `/version` returns that value; for non-release builds (per-merge `:latest`/`:<sha>`) it reports the **short SHA** rather than a bare "unknown", so every image is traceable.
- The help footer reads its version from `/version` at load time (replacing the dead `button_status` firmware/device block), touching script logic only — the visible markup is unchanged. The hardcoded string remains only as a fetch-failure fallback.
- Per-merge build behavior is otherwise unchanged (`:latest` + `:<sha>` still published).

This slice needs no new workflow action — the version can be passed manually as a build arg to verify the whole pipe.

Respects ADR-016, ADR-002 (Watchtower), ADR-003 (plain-HTML frontend). nginx already proxies `/version` to the backend, so the static UI image's fetch resolves.

## Acceptance criteria

- [ ] Building the API image with a version build-arg results in `/version` returning that exact version
- [ ] A build with no version provided reports the short commit SHA (never a bare "unknown")
- [ ] The running image carries the version as an OCI image label, visible via image inspection
- [ ] The help footer displays `V <version>` sourced from `/version`; the dead firmware/device code is removed, not layered over
- [ ] The footer's visible markup/layout is unchanged (only script logic differs)
- [ ] Contract test covers `/version` returning the env value and the SHA/fallback behavior (mirrors the existing button-endpoint contract tests)
- [ ] (Optional, lower priority) e2e test confirms the footer renders the version from `/version`
- [ ] Per-merge `:latest`/`:<sha>` publishing still works

## Blocked by

None - can start immediately

nana-acorn:
Closes #229. First tracer-bullet slice of #228 (image versioning).

## What this delivers
Every built image now carries a version and self-reports it end-to-end:

- **`Dockerfile.api`** — `AQ_APP_VERSION` build arg → env var + `org.opencontainers.image.version` OCI label.
- **`docker-build.yml`** — per-merge builds stamp the **short commit SHA** into `AQ_APP_VERSION` (new `SHORT_SHA` var), so non-release images are traceable, never a bare "unknown".
- **`/version`** — hardened to return `"unknown"` for an empty *or* unset `AQ_APP_VERSION` (an empty build-arg previously yielded `""` → a blank `V` in the footer).
- **`help.html`** — the footer now `fetch("/version")` and shows `V <version>`, replacing the dead `button_status` firmware/device block (which depended on fields the endpoint never returned). Script logic only — visible markup unchanged; hardcoded value kept as a fetch-failure fallback.

## Tests (TDD, contract seam)
`tests/contract/test_version_endpoint.py` — built one-test-at-a-time:
1. `/version` returns the configured version when set
2. `/version` → `"unknown"` when unset
3. `/version` → `"unknown"` (not `""`) when empty  ← drove the endpoint change

`pytest tests/contract/test_version_endpoint.py` → 3 passed. Full collectable suite: 495 passed, 0 new failures (6 pre-existing wifi/env failures and 2 hardware-import collection errors confirmed present on clean main, unrelated to this change).

## Deferred / out of scope
- e2e footer-rendering test (`seam:e2e`) — deferred per the issue (optional/lower priority).
- The `cut-version` workflow action, version minting, and Releases are #230.

Design per ADR-016.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -21,8 +21,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Set lowercase repo
-        run: echo "REPO_LOWER=$(echo "$GITHUB_REPOSITORY" | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV
+      - name: Set build vars
+        run: |
+          echo "REPO_LOWER=$(echo "$GITHUB_REPOSITORY" | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV
+          echo "SHORT_SHA=${GITHUB_SHA::7}" >> $GITHUB_ENV
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
@@ -44,6 +46,8 @@ jobs:
           file: docker/Dockerfile.api
           platforms: linux/amd64,linux/arm64
           push: true
+          build-args: |
+            AQ_APP_VERSION=${{ env.SHORT_SHA }}
           tags: |
             ghcr.io/${{ env.REPO_LOWER }}-api:latest
             ghcr.io/${{ env.REPO_LOWER }}-api:${{ github.sha }}
--- a/aquila_web/main.py
+++ b/aquila_web/main.py
@@ -659,7 +659,7 @@ async def health_check():
 
 @app.get("/version")
 async def version_check():
-    return {"version": os.getenv("AQ_APP_VERSION", "unknown")}
+    return {"version": os.getenv("AQ_APP_VERSION") or "unknown"}
 
 
 @app.get("/results")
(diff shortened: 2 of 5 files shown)


nana-acorn:
ℹ️ Heads-up: PR #234 (#230, cut-version) is **stacked on this branch**. After this merges to `main`, #234 needs a rebase onto `main` + base retarget. Merge this one first.

PR #269, opened 2026-06-30
#267 History shows profile name instead of file-path id

### LINKED ISSUE #267: History page shows profile file path/slug instead of the profile name
## Summary
On the **History** page, the **Profile** column shows the profile's file-path id (e.g. `local\A3_Invalid_Temp.json`) instead of the profile's human-readable name (`A3 Invalid Temp`). The path prefix (`local/`), the `.json` extension, and the underscores-for-spaces in the filename should never be shown to the user.

## Steps to reproduce
1. Create/select a profile named `A3 Invalid Temp`.
2. Run it (dev simulate).
3. Go to **History**.

**Expected:** Profile column shows `A3 Invalid Temp`.
**Actual:** Profile column shows `local\A3_Invalid_Temp.json`.

## Root cause
`/profile/select` stores the profile **id** (the relative path emitted by `GET /profiles`, e.g. `local/A3_Invalid_Temp.json`) in `selected_profile`. When a run completes, `_simulate_run` writes that id verbatim into the history entry's `profile` field (`aquila_web/main.py:585`). The History page (`static/history.js:135`) renders `entry.profile` directly, so the user sees the path/slug.

The profile's real display name lives in the profile JSON's `name` field (`GET /profiles` already exposes it as `name`/`display_name`).

## Proposed fix
Resolve the profile id to its display name (the JSON `name`) when building the history entry, so `profile` stores `A3 Invalid Temp`. Apply the same resolution to the `/history/append` write path defensively (idempotent for values already equal to a name).

## Out of scope
- History entries already written with the old path value (only new runs are corrected).
- The `run_complete` event payload's `profile` field.

jackhu23:
Closes #267

## Problem
On the **History** page, the **Profile** column showed the profile's file-path id (e.g. `local\A3_Invalid_Temp.json`) instead of the profile's name (`A3 Invalid Temp`) — leaking the `local/` directory, the `.json` extension, and the filename's underscores-for-spaces.

## Root cause
`POST /profile/select` stores the profile **id** (the relative path emitted by `GET /profiles`) in `selected_profile`. On run completion, `_simulate_run` wrote that id verbatim into the history entry's `profile` field, and `static/history.js` renders it directly.

## Fix
- New `_resolve_profile_display_name(profile_ref)` helper in `aquila_web/main.py` — resolves an id to the profile JSON's `name` (falling back to `title`, then a path-free stem). Idempotent for values already equal to a display name.
- Used it in `_simulate_run` (the buggy write path) and `append_history` (defensive).
- Production change is **+42 / −2 lines**; the rest is tests + spec.

## Tests
- `unit_tests/test_history_profile_name.py` — 5 unit tests (path-id → name, no path/extension/underscore artifacts, idempotency, empty → `--`, missing-file fallback).
- `tests/contract/test_history_endpoints.py` — new contract test that reproduced the exact bug before the fix and now asserts `A3 Invalid Temp`.
- All 16 pass. Full suite introduces no new failures (pre-existing wifi/optics env failures are unrelated).

Run: `pytest unit_tests/test_history_profile_name.py tests/contract/test_history_endpoints.py -v`

## Out of scope
- History entries written **before** this fix keep their old value (only new writes are corrected).
- The `run_complete` event payload's `profile` field.

Spec: `specs/backend/spec_history_profile_display_name.md`

### DIFF
--- a/aquila_web/main.py
+++ b/aquila_web/main.py
@@ -103,6 +103,46 @@ def _profile_rox_unavailable(profile_name: str | None) -> bool:
             continue
     return False
 
+def _resolve_profile_display_name(profile_ref: str | None) -> str:
+    """Resolve a profile reference to its human-readable display name.
+
+    ``profile_ref`` is normally the id stored by ``/profile/select`` — the
+    relative path emitted by ``GET /profiles`` (e.g. ``local/A3_Invalid_Temp.json``).
+    History must show the profile's ``name`` (e.g. ``A3 Invalid Temp``), never the
+    ``local/`` prefix, the ``.json`` extension, or the filename's underscores
+    (issue #267). Idempotent when given a value that is already a display name.
+    See specs/backend/spec_history_profile_display_name.md.
+    """
+    if not profile_ref:
+        return "--"
+    profile_dir = resolve_profile_dir()
+    # 1. Normal path: treat the ref as the relative-path id and read its JSON name.
+    try:
+        candidate = profile_dir / profile_ref
+        if candidate.is_file():
+            data = json.loads(candidate.read_text())
+            name = data.get("name") or data.get("title")
+            if name:
+                return str(name)
+    except Exception:
+        pass
+    # 2. Otherwise match by name/stem/filename/id across the profile tree.
+    try:
+        if profile_dir.exists():
+            for path in profile_dir.rglob("*.json"):
+                try:
+                    data = json.loads(path.read_text())
+                except Exception:
+                    continue
+                name = data.get("name") or data.get("title") or path.stem
+                rel = str(path.relative_to(profile_dir))
+                if profile_ref in (name, path.stem, path.name, rel):
+                    return str(name)
+    except Exception:
+        pass
+    # 3. Fallback: strip directory and extension so a path is never shown.
+    return Path(profile_ref).stem
+
 def _all_bundled_filenames() -> set[str]:
     profile_groups_path = BASE_DIR / "config_files" / "profile_groups.json"
     try:
@@ -546,7 +586,7 @@ async def _simulate_run(profile_name: str) -> None:
     history = _load_history()
     history.append({
         "timestamp": datetime.now().strftime("%Y-%m-%d %H:%M"),
-        "profile": profile_name,
+        "profile": _resolve_profile_display_name(profile_name),
         "run_name": run_name,
         "result": detected_summary,
         "graph_path": f"/plots/{plot_filename}",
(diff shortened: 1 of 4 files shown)


jackhu23:
Thanks for the thorough reviews — addressed the dots concern in 044151f.

**Fallback fix:** step 4 now uses `Path(ref).name.removesuffix(".json")` instead of `Path(ref).stem`, so an unresolvable reference keeps dots that are part of the name:
- `"Sample 3.2 Panel"` → `"Sample 3.2 Panel"` (was `"Sample 3"`)
- `"local/Cycle 2.5.json"` → `"Cycle 2.5"`

Added a unit test (`test_fallback_preserves_dots_in_name`) and updated the spec. Full file now 18 tests, all green:
`pytest unit_tests/test_history_profile_name.py tests/contract/test_history_endpoints.py -v` → 18 passed.

**On the other notes (left as-is, all non-blocking):**
- *Path-traversal in `append_history`* — agreed it's a real (very low severity, local kiosk, info-leak only) sharp edge, but out of scope for #267 and consistent with the endpoint's existing loose `results_path` handling. Worth a dedicated boundary-hardening pass.
- *Underscore-vs-space in fallback* — only reachable when no file matches; can't be fixed without reading the file. Acceptable degradation, documented in the spec.
- *`run_complete` event / `/state` raw id* — explicitly out of scope; those don't feed the History page.
- *PR description "wifi/optics" label* — correct, that label was machine-specific; the accurate statement is "pre-existing, environment-dependent failures unrelated to this change, zero new failures introduced."

NicoleCornell8 (approved):
Approved. Clean, tightly-scoped fix — resolving the profile id to its display name at write-time is the right call, and the test coverage is strong (17 tests pass locally, including a contract test that genuinely reproduces the bug and edge cases for idempotency, missing files, and dot-preservation).

A few non-blocking follow-up notes:
1. **Minor security** — step 1 of `_resolve_profile_display_name` reads whatever `profile_ref` points at, and that value comes from the user-controlled `/profile/select` payload. A crafted ref (`../../some.json`) could surface an arbitrary JSON file's `name`/`title`. Low severity (local operator, only a string leaks, non-JSON is swallowed), but consider asserting `candidate.resolve().is_relative_to(profile_dir.resolve())`.
2. **Live display** — `/button_status` still returns the unresolved `selected_profile`; if any run UI reads `profile` from there, the raw id may still flash during a run. Possible follow-up.
3. **`rglob` cost** — the defensive `append_history` path walks the full profile tree when step 1 misses. Negligible now; noting in case the tree grows.

None block merge.

Issue #288, opened 2026-07-02
Device: optics_readings capture at run completion


NicoleCornell8:
## What to build

After `run_complete` is enqueued, read the **exact optics file** `process_run()` just consumed, gzip it, compute `sha256` / `raw_bytes` / `line_count` and `expected_lines` (`cycle_count × 480`) / `complete`, and enqueue an **`optics_readings`** event on the existing `events` outbox carrying the **same `run_timestamp`** as `run_complete`. An **aborted** run emits whatever was captured labeled `complete=false, aborted=true`; an aborted run with **no** capture emits **no** event (so the coverage view shows it as genuinely missing).

Source: PRD stories 9, 11–17; ADR-0007.

## Acceptance criteria

- [ ] `optics_readings` enqueued right after `run_complete`, same `run_timestamp`
- [ ] Payload matches the frozen contract (AcornGenetics/acorn-analytics#45); blob gzipped + base64'd
- [ ] `complete` computed from `line_count` vs `cycle_count × 480`
- [ ] Aborted-with-partial → `complete=false, aborted=true`; aborted-no-capture → no event
- [ ] Device tests for the completeness + abort branches

## Blocked by

- AcornGenetics/acorn-analytics#45 (frozen contract + completeness constant)
- AcornGenetics/aquilla-main#287 (shared run_timestamp)

PR #290, opened 2026-07-02
feat(sync): size-aware Sync + size guard + chunk split (#289)

### LINKED ISSUE #289: Device: size-aware Sync + size guard + chunk split
## What to build

Make **Sync size-aware**: send an `optics_readings` event in its **own POST** (or a byte-capped batch) so a large optics payload never pushes a batch past the SQS limit; add a **size guard** that refuses to enqueue/flush an oversized single event (**fail loud**, never silently drop/truncate); and above the inline limit, **split** the gzipped log into ordered chunks that share one `sha256`.

Sizing reality: current logs are ≤230 KB base64 (under the 256 KB ceiling), so chunking is a guard for future longer protocols, not an everyday path. Source: PRD stories 18–20.

## Acceptance criteria

- [ ] An optics event is sent alone / in a batch capped by bytes under the SQS limit
- [ ] An oversized single event fails loudly (no silent drop or truncation)
- [ ] Above the inline limit, the log splits into ordered chunks sharing one `sha256`
- [ ] Sync tests for batching, the size guard, and the split

## Blocked by

- AcornGenetics/aquilla-main#288 (optics_readings capture)

jackhu23:
Closes #289.

## What & why

Makes Sync **size-aware** so a large `optics_readings` payload (from #288, ~230 KB gzipped) never pushes a POST past the **SQS 256 KB message limit** — which would get the message silently rejected. Three complementary defenses, none of which drops or truncates data:

| Mechanism | What it does |
|---|---|
| **Byte-capped batching** | Pack pending events into batches each ≤ the cap; one POST per batch. A large optics event lands **alone** in its own POST. |
| **Size guard** | A single event that can't fit even alone is **quarantined** — left pending, logged loudly, never truncated. Healthy events keep flowing. |
| **Chunk split** | `split_log()` slices an over-limit gzipped blob into ordered chunks sharing one `sha256` (reconstructs exactly). Built ready for #288's producer to call; not yet wired to a capture path. |

## Changes

- **New** `aquila_web/sync_batching.py` — pure, no IO: `batch_events`, `partition_oversized`, `split_log`, `event_size_bytes`.
- `aquila_web/sync.py` — `sync_pending_events()` becomes a per-batch POST loop: quarantine oversized, mark-synced **per batch** after success, `break` on network error so earlier batches stay synced and later ones stay pending. New `AQ_SYNC_MAX_MESSAGE_BYTES` env knob for the cap.
- Spec: `specs/backend/spec_size_aware_sync.md`.

## Design decisions

- **Guard quarantines rather than aborts** the whole flush — aborting would let one poison event block every healthy event behind it (its own silent failure). A backstop must not take the queue hostage.
- **Batching + guard wired live now** (generic, byte-based, doesn't need #288). **`split_log` built but not wired** — it consumes #288's gzip blob + `sha256`, which don't exist yet.

## Tests

- `unit_tests/test_sync_batching.py` — 7 pure-logic tests (packing, sent-alone, order, guard partition, single-chunk, multi-chunk reconstruct).
- `tests/unit/test_background_sync.py` — +3 flush tests (separate POSTs over cap, quarantine, partial-sync on later-batch failure). Existing 5 sync tests stay green. **15 pass.**

## Open question (see spec §9)

Does #288 split into chunk *events* at enqueue time using `split_log`, or does Sync split at flush? This PR assumes producer-side split — confirm when #288 lands.

## Blocked-by note

Runtime end-to-end (real optics capture → sync) depends on #288; this PR is built and tested against the documented event shape with synthetic large events.

### DIFF
--- a/aquila_web/local_db.py
+++ b/aquila_web/local_db.py
@@ -42,6 +42,14 @@ def init_local_db() -> None:
             )
             """
         )
+        # Additive migration for existing devices: quarantine column for events
+        # too large to Sync (#289). Kept out of the flushable set without dropping
+        # the row, so a poison event is handled once and never re-fetched.
+        columns = {row["name"] for row in connection.execute("PRAGMA table_info(events)")}
+        if "quarantined_at" not in columns:
+            connection.execute("ALTER TABLE events ADD COLUMN quarantined_at TEXT")
+        if "quarantine_reason" not in columns:
+            connection.execute("ALTER TABLE events ADD COLUMN quarantine_reason TEXT")
 
 
 def enqueue_event(event_type: str, payload: dict[str, Any], device_id: str | None = None) -> int:
@@ -63,7 +71,7 @@ def get_pending_events(limit: int = 100) -> list[dict[str, Any]]:
             """
             SELECT id, event_type, payload, device_id, created_at
             FROM events
-            WHERE synced_at IS NULL
+            WHERE synced_at IS NULL AND quarantined_at IS NULL
             ORDER BY id ASC
             LIMIT ?
             """,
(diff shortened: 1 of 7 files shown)


jackhu23:
Pushed `c110523` addressing the review concerns.

**1. Quarantine log spam / slot starvation — fixed.** Oversized events were left pending, so they were re-fetched, re-serialized, and re-logged (ERROR) every flush and permanently held a batch slot. Added additive `quarantined_at` / `quarantine_reason` columns; `get_pending_events` now excludes quarantined rows and the guard marks a poison event quarantined **once**. The row is retained (never dropped/truncated) and discoverable via `get_quarantined_events`.

**2. Coarse 4 KB envelope reserve — fixed.** Replaced the fixed `ENVELOPE_OVERHEAD_BYTES = 4096` with the *measured* `{device_id, events:[...]}` wrapper (`envelope_overhead_bytes` / `max_batch_bytes`, device-aware). An event just under the 256 KB ceiling now syncs instead of being needlessly quarantined.

**3. `split_log` wiring — intentionally deferred.** Per discussion, #288 is being built concurrently and owns the producer-side split, so `split_log` stays as the tested, ready-to-consume function. Spec §9 updated to reflect this rather than leaving it as an open ambiguity.

**Tests:** new `unit_tests/test_local_db_quarantine.py` (additive-migration idempotency + quarantine seam), plus flush tests for no-reprocess-on-next-flush and near-ceiling-still-syncs, plus pure envelope-overhead tests. **23 pass** across the three suites; existing sync behaviors stay green.

jackhu23:
Pushed `75f9786` addressing review round 2.

**Blocker — 2-byte item separator undercount — fixed.** `json.dumps` uses a 2-byte `", "` between array items; the reserve only counted 1, so a maximally-packed batch of many small events could still serialize past the 256 KB ceiling. Fix: the separator is now counted **per-event inside `batch_events`** as it packs, so the cap reserves only the measured wrapper — a packed batch can no longer overflow. This also removed the pessimistic upfront `max_events` reserve that the off-by-one lived in. Added an invariant test (a packed 200-event batch serializes under the ceiling) and corrected the test that had codified the off-by-one.

**Lower-priority hardening — both done.** `AQ_SYNC_MAX_MESSAGE_BYTES` is now validated: non-numeric or ≤ the envelope overhead falls back to the real ceiling with a warning, instead of crashing the flush or driving the cap negative and irreversibly mass-quarantining every event. Parametrized test covers `non-numeric / 5 / -100 / 0`.

**Tests:** 27 pass across the three suites.

Issue #287, opened 2026-07-02
Device: canonical shared run_timestamp on run_complete


NicoleCornell8:
## What to build

Set **one canonical `run_timestamp` per Run** and include it on the `run_complete` event payload, so run identity no longer depends on enqueue-time `created_at`, and the forthcoming `optics_readings` event can reuse the same value → both derive the same `run_id` cloud-side (`uuid5(device_id : run_timestamp)`).

Source: `docs/prd/optics-log-capture-and-warehouse.md` (story 10), `acorn-analytics/docs/adr/0007`.

## Acceptance criteria

- [ ] A single canonical `run_timestamp` is computed once per Run
- [ ] `run_complete` payload carries `run_timestamp`
- [ ] Existing `run_complete` flow and tests still pass

## Blocked by

None - can start immediately.

PR #293, opened 2026-07-02
feat(optics): capture optics_readings at run completion (#288)

### LINKED ISSUE #288: Device: optics_readings capture at run completion

jackhu23:
Closes #288.

## What this does
After `run_complete` is enqueued, the device captures the exact optics log the run consumed into an **`optics_readings`** event on the same SQLite outbox, carrying `run_complete`'s canonical `run_timestamp` (#287). Payload matches the frozen device↔cloud contract (AcornGenetics/acorn-analytics#45).

## Key decision — completeness formula
#288's body says `expected_lines = cycle_count × 480`, but the **frozen contract #45 overrides that** and explicitly says "do not hardcode 480 — #288 must read reads_per_cycle from the profile." This PR follows #45:

`expected_lines = read_passes × reads_per_cycle × 60`, where `read_passes = cycle_count + 1` (baseline + one per cycle), `reads_per_cycle = 8`, `SAMPLES_PER_BLINK = 60`.

Verified against the device: `capture_blink` writes 60 samples/blink, `read_wells` fires 8 blinks/pass. A real 40-cycle run = 41 passes × 480 = 19,680 rows — #288's `40×480` would flag a healthy run as `complete=false`.

## Changes
- **`aquila_web/optics_readings.py`** (new) — pure builder: `sha256` of raw bytes, `raw_bytes`, data-row `line_count`, gzip+base64 `data_b64` (single chunk), strict `complete = line_count == expected_lines`. Returns `None` for aborted-with-no-capture → **no event**.
- **`/events/optics_readings`** endpoint + **`state_requests.emit_optics_readings()`** — device hands the optics path to the local backend, which reads/gzips/hashes/enqueues.
- **`state_run_assay.run()`** — counts optical passes; emits after `emit_run_complete` (success) and on the abort path (`aborted=True`).
- **`_simulate_run`** — emits for the dev/demo path too.
- **`tests/fixtures/optics/{sample,partial}.log`** — shared fixtures, byte-identical to acorn-analytics (SHA-256 verified), LF-pinned via `.gitattributes` + a `.gitignore` negation.

## Acceptance criteria
- [x] `optics_readings` enqueued after `run_complete`, same `run_timestamp`
- [x] Payload matches frozen contract; blob gzipped + base64'd
- [x] `complete` from `line_count` vs `expected_lines`
- [x] Aborted-with-partial → `complete=false, aborted=true`; aborted-no-capture → no event
- [x] Device tests for the completeness + abort branches

## Testing
`pytest tests unit_tests` — **676 non-e2e tests pass**, including 11 new. The e2e failures are pre-existing environmental (Playwright `page` fixture not installed); e2e is verified locally against a live backend.

## Reviewer notes
- The `state_run_assay.py` wiring is hardware-coupled and (like the existing `run_complete` wiring) has no automated test — the unit-tested builder/endpoint/emit sit underneath it.
- `_simulate_run` labels its consumed file `complete=true` (a simplification for the demo path, not specified by #288).


NicoleCornell8 (commented):
## Code review — optics_readings capture (#288)

Overall this is a **clean, faithful implementation** of the frozen contract (acorn-analytics#45) and ADR-0007. The real-run path derives `expected_lines` correctly from a runtime optical-pass count rather than a hardcoded value, both the completion and abort paths emit with the **shared `run_timestamp`**, and the aborted-no-capture case correctly emits nothing. Good, targeted test coverage. Nothing here is blocking — findings below are follow-ups and nits.

### What's right
- **Shared `run_timestamp` (Option A)** on both `run_complete` and `optics_readings` → same `run_id` cloud-side. ✅
- **`expected_lines` = `read_passes × reads_per_cycle × 60`**, with `read_passes` from the runtime `_optics_pass_count` (`state_run_assay.py`) — the #45 correction, not a hardcoded 480. ✅
- **Aborted handling both ways**: abort path emits `aborted=True` (partial), completion emits `aborted=False`; aborted-with-no-capture → `None` → no event. ✅
- Reuses the `events` outbox; whole-file gzip + sha256 + `chunk_count=1` per ADR-0007. ✅
- Tests cover the frozen payload, partial/aborted, the None case, the formula, line-counting, and the endpoint enqueue with `run_timestamp`.

### Findings

**1. [medium] `READS_PER_CYCLE = 8` is hardcoded.** The #45 correction is that `reads_per_cycle` is *profile-dependent*; here it's a module constant coupled to `read_wells`' blink pattern (rox 0–3 + fam 2–5), not the profile's optics config. Correct for today's read pattern, but a profile with a different optics configuration would silently miscompute `complete`. This is a narrower cousin of the "hardcoded 480" problem. Suggest either documenting that `READS_PER_CYCLE` is coupled to `read_wells` (and asserting it there), or deriving it from the profile's optics steps. (The runtime `_optics_pass_count` already handles `read_passes` correctly — only blinks-per-pass is hardcoded.)

**2. [low] No size guard; `chunk_count` always 1.** The size-aware Sync + fail-loud guard (#289) was reverted (#291), so an optics event whose gzip+base64 exceeds the ~256 KB SQS limit is currently unprotected. Fine for today's ~230 KB runs — flagging the dependency so a longer protocol doesn't silently exceed the limit. #289 needs redoing before that can happen.

**3. [low] A non-aborted run with an empty/missing optics file emits a present-but-empty event.** `build_optics_readings` returns `None` only when `aborted and line_count == 0`. A *non-aborted* run with a missing file yields a row with `line_count=0`, `sha256(b"")`, `complete=false` — which shows as "present but empty" rather than "missing" in `v_runs_missing_optics`. Pathological (a normal run always has the file `process_run` consumed), but consider returning `None` whenever `raw` is empty regardless of `aborted`, so a missing capture is always genuinely missing.

**4. [nit] Parameter shadows the module function.** In `build_optics_readings`, the `expected_lines: int` parameter shadows the module-level `expected_lines()` function. It works, but a rename (e.g. `expected_lines_count`) removes a footgun if someone later calls the function inside `build`.

**5. [nit] Sim path reads the file twice.** In `_simulate_run`, `count_data_lines(optics_path)` decodes the file, then `build_optics_readings` reads + decodes it again. Minor (sim only) — could pass the raw bytes through.

Nice work — #1 is the one I'd genuinely consider addressing (or explicitly documenting) before this rides on a non-standard profile.

NicoleCornell8 (approved):
Approve. Well-engineered and well-tested — verified the core claims independently:

- **Completeness math has one source of truth.** `thermal_parser` yields the `optics` action → executor dispatches 1:1 to `read_wells` → `count_optics_passes` counts that same action, so planned and runtime pass counts can't drift. `READS_PER_CYCLE` derived from `OPTICS_READ_PLAN` (=8) is sound.
- **Path confinement is effective.** Device sends a relative `logs/optics/*.log`; endpoint resolves against BASE_DIR and requires `.resolve()` under OPTICS_LOG_DIR — real payloads pass, the device.env exfil path is rejected (400, no event), with a regression test.
- **Sync-handler-in-threadpool is safe.** `local_db._connect()` opens a fresh WAL connection per call, so no SQLite connection is shared across threads.
- **Abort semantics are honest.** Using planned passes for `expected_lines` correctly reports `complete=false` on a mid-run abort. 15/15 new unit tests pass locally.

Minor, non-blocking:
1. `expected_lines` the function shadows the `expected_lines` param/field — consider renaming one for readability.
2. `_simulate_run` reads the optics file twice (`count_data_lines` then `build_optics_readings`); the builder already computes `line_count`.
3. Strict `complete == line_count` is brittle by design (frozen contract) — expect occasional benign `complete=false` from off-by-one rows.

None block merge.

PR #300, opened 2026-07-09
feat(events): emit per-Well sample names in run_complete (#296)

### LINKED ISSUE #296: [qPCR Analysis A3] Emit per-Well Sample Names in the run_complete event
## What to build

Today the operator's tube names live only in device-local history (`current_tube_names`) and are NOT included in the `run_complete` event, so they never reach the Warehouse. Add per-Well sample names to the emitted `run_complete` event, keyed to well 1-4.

Part of the qPCR Run Analysis redesign (acorn-internal-app PRD 0002, Section A) and ADR-0005.

## Acceptance criteria

- [ ] The `run_complete` event payload includes per-Well sample names keyed to well 1-4
- [ ] The event still carries `run_name` (no regression)
- [ ] Default tube names ("Tube 1".."Tube 4") are emitted when the operator hasn't renamed
- [ ] Event-seam test asserts the sample names appear in the emitted payload

## Blocked by

None - can start immediately

jackhu23:
Closes #296.

## What & why

The operator's tube names lived only in device-local state (`current_tube_names`) and were never included in the `run_complete` event, so they never reached the Warehouse. This adds a `sample_names` field to the emitted `run_complete` payload, keyed to Wells 1–4.

```json
"sample_names": {"1": "Tube 1", "2": "Tube 2", "3": "Tube 3", "4": "Tube 4"}
```

Part of the qPCR Run Analysis redesign (PRD 0002 Section A, ADR-0005).

## How

- New `_sample_names_by_well()` helper normalizes `current_tube_names` and keys it to Wells 1–4 (string keys, matching the integer well numbering in `calls[]`), defaulting to `"Tube 1".."Tube 4"`.
- Wired into **both** run_complete emit paths for parity: the `/events/run_complete` endpoint (real hardware) and `_simulate_run` (sim/e2e).
- `CONTEXT.md` `Event` glossary entry updated to document the new field.

## Acceptance criteria

- [x] Payload includes per-Well sample names keyed to well 1–4
- [x] Event still carries `run_name` (no regression — explicitly asserted)
- [x] Default tube names emitted when the operator hasn't renamed
- [x] Event-seam test asserts the sample names appear in the emitted payload

## Testing

Built test-first (red→green). All 9 tests in `tests/unit/test_run_complete_event.py` pass, including two new event-seam tests (default + operator-renamed). Verified via `git stash` that the only other suite failures on this Windows/headless env (e2e browser tests, `test_result_flow`/`test_seams`) are pre-existing and unrelated — zero regressions introduced.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -82,7 +82,7 @@ The fractional PCR cycle at which a Well's fluorescence crosses the detection th
 The fraction of Calls with outcome `Inconclusive`, grouped by Protocol. The primary analytics metric for v1. Denominator excludes `ROX Unavailable` calls.
 
 **Event**
-A structured JSON record enqueued in the local SQLite database (`data/db/app.db`) when a Run completes. Event type: `run_complete`. Payload contains: protocol name, run name, run timestamp, duration, and all 8 Well × Channel Calls with Cq values. Events queue offline and flush on the next successful Sync.
+A structured JSON record enqueued in the local SQLite database (`data/db/app.db`) when a Run completes. Event type: `run_complete`. Payload contains: protocol name, run name, run timestamp, duration, per-Well sample names (keyed to Wells 1–4, defaulting to "Tube 1".."Tube 4"), and all 8 Well × Channel Calls with Cq values. Events queue offline and flush on the next successful Sync.
 
 **Sync**
 The background process (asyncio task in the FastAPI app, 15-minute interval) that batches pending Events from the local SQLite queue and POSTs them to the AWS Ingest Endpoint. Also triggered on WiFi reconnect. On success, marks events `synced_at` in SQLite. Events accumulate indefinitely if offline.
--- a/aq_lib/state_requests.py
+++ b/aq_lib/state_requests.py
@@ -62,17 +62,24 @@ def mark_results_ready(path: str | Path) -> None:
     except requests.exceptions.RequestException as e:
         logger.exception("Error marking results ready: %s", e)
 
-def log_history(profile, run_name, results_path, graph_path=None):
-    url = f"{BACKEND_URL}/history/append"
-    resolved_results_path = None
-    tube_names = None
+def get_tube_names():
+    # Snapshot the operator's current tube names once, so history and the
+    # run_complete event can be fed the SAME labels captured with the run
+    # rather than each re-reading the mutable server global (#296). Returns
+    # None on failure; callers fall back to their own defaults.
     try:
         response = requests.get(f"{BACKEND_URL}/tube_names", timeout=5)
         if response.ok:
-            data = response.json()
-            tube_names = data.get("names")
+            return response.json().get("names")
     except requests.exceptions.RequestException:
-        tube_names = None
+        pass
+    return None
+
+def log_history(profile, run_name, results_path, graph_path=None, tube_names=None):
+    url = f"{BACKEND_URL}/history/append"
+    resolved_results_path = None
+    if tube_names is None:
+        tube_names = get_tube_names()
     if results_path:
         base_dir = Path(get_src_basedir())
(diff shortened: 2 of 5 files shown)


NicoleCornell8 (commented):
### Finding: `sample_names` reads a mutable global at emit time instead of being captured with the run

**Severity: medium** (silent, timing-dependent)

`_sample_names_by_well()` sources the labels from the live `current_tube_names` global at the moment the event is enqueued (`main.py:360`, called at `main.py:883`). Every other run-scoped field is captured *with* the run and passed through the request — `run_name` and `run_timestamp` come from `self.run_name` / `self.run_timestamp` at the call site (`state_run_assay.py:306`), and `calls`/`result` come from the run's `results_path`. `sample_names` is the only field that reads "whatever the global says right now."

**Failure scenario:** if `current_tube_names` changes between the run finishing and the event being enqueued, the event ships the wrong labels alongside the correct run. Two realistic windows:
1. **Reset button** → `/results/clear` resets the global to `"Tube 1".."Tube 4"` (`main.py:840`), so the operator's real names are silently replaced with defaults.
2. **Multiple clients** → one browser edits tube names for the next run while another's `run_complete` is being emitted (the documented multi-client flow, `docs/testing/tests.md:924`).

It fails silently (plausible-looking default names), which is the worst mode for a feature whose purpose is preserving what the operator typed.

**Why this is a real gap, not theoretical:** `log_history` already solves the same problem the right way — it snapshots the names via `GET /tube_names` and stores them **by value** (`state_requests.py:65-88`). The new event code reads the live global instead, so the same run can end up with *different* tube names in history vs. the event, violating the documented invariant "History entry `tube_names` match labels shown at run completion" (`docs/testing/tests.md:480`).

### Suggested fix

Capture the names once at run completion and thread them through the request, mirroring how `run_name`/`run_timestamp` already flow — don't read the global at enqueue time.

1. `_sample_names_by_well(names: list | None = None)` — accept an explicit snapshot, fall back to the global for the sim/legacy path.
2. Add `tube_names: Optional[list] = None` to `_RunCompleteEventRequest`.
3. Handler uses `_sample_names_by_well(req.tube_names)` instead of the bare global.
4. `emit_run_complete(...)` sends `tube_names` in the POST body.
5. Lift the `GET /tube_names` out of `log_history` into a shared `get_tube_names()` snapshot taken once at completion, passed to **both** `log_history` and `emit_run_complete` — so the two records match by construction rather than each doing its own lookup.

```
        get_tube_names()      # snapshot once, at completion
        /            \
  log_history      emit_run_complete
  (history)        (event)          # both fed the SAME snapshot
```

The two new event-seam tests would then send `tube_names` in the request rather than relying on the server reading the global (the `/results/clear`-then-expect-defaults test in particular encodes the buggy behavior).

### Also fine to note
- No crash risk: `_normalize_tube_names` always returns exactly 4 entries, so `names[well-1]` for wells 1–4 can't go out of bounds.
- Both emit paths (sim + real) correctly get the field — good parity.
- Minor: `_sample_names_by_well` could iterate the normalized list directly (`{str(i+1): n for i, n in enumerate(names)}`) instead of the `range` + `well-1` indexing.

_Reviewed via /code-review (high-effort, recall-biased)._

NicoleCornell8 (approved):
Approved. Clean, well-tested change — the single-snapshot design threads run labels through both history and the run_complete event correctly, and the success path is provably consistent.

One minor non-blocking note for later: if the one-time `get_tube_names()` snapshot fails (returns `None`), the two records each re-read independently instead of sharing a saved copy, so a concurrent tube-rename could produce mismatched labels. Extremely rare (needs a localhost call to fail *and* a simultaneous edit on another client) and never crashes or loses data. Worth a small future cleanup: capture once, reuse the copy, don't re-grab.

PR #313, opened 2026-07-10
fix(run): countdown survives Reset + re-run; resolve estimate by name recursively (#312)

### LINKED ISSUE #312: Countdown timer regresses to stopwatch after run + Reset + re-run
## Summary

The countdown timer (from #110 / #121) regresses to the stopwatch after a run + Reset + re-run. If a profile has an estimated completion time, the first run correctly shows the **countdown** ("Time Remaining"). But after that run completes, the operator hits **Reset** and runs the same profile again — the Run screen shows the **stopwatch** ("Elapsed Time") instead of the countdown.

Reported against the **beer spoilage** profile (which has an estimate set).

## Steps to reproduce

1. Select a profile that has an estimated completion time (e.g. beer spoilage).
2. Run it — the timer correctly counts **down** ("Time Remaining"). ✅
3. Let the run complete.
4. Hit **Reset**.
5. Run the profile again.
6. **Bug:** the timer shows the count-up **stopwatch** ("Elapsed Time") instead of the countdown. ❌

## Root cause

The countdown-vs-stopwatch decision depends entirely on a single client-side cache, `cachedEstimateSeconds`, which is only populated when a profile is **selected through the UI** — it is never re-derived at run start.

- `cachedEstimateSeconds` is written only inside `loadProfileLabels()` (`aquila_web/static/script.js:146`), which is called only from selection paths: the dropdown `change` handler (`script.js:1394`) and the initial page-load restore (`script.js:1350`).
- `beginTimerMode()` blindly trusts that cache when the run transitions to `running` (`script.js:225-231`). `notifyRun()` re-syncs `/profile/select` to the server before running (`script.js:798`) but never refreshes the estimate cache.
- **Reset clears the profile server-side.** The Reset button → `resetRunScreen()` → `POST /run/complete/ack` → `_set_selected_profile(None)` (`aquila_web/main.py:846-852`, added for #275 so the ready screen returns to "Select a profile" with no manual reset). The authoritative `selected_profile` is now null (and persisted to disk).

After a completed run + Reset, the server profile is cleared and the ready screen resets. When the profile is subsequently restored from server/persisted state (page reload, or ready-screen reconciliation) **without going back through `loadProfileLabels()`**, `cachedEstimateSeconds` is stale/`null` while the profile still looks selected. The next run's `beginTimerMode()` reads `null` and falls back to the stopwatch. The first run works because the operator's manual selection warmed the cache; the second run reuses the profile without re-selecting, so the cache is never re-warmed.

### Note on exact trigger

Two candidate paths lead to the null cache — a page-reload cold cache, or an async race where `loadProfileLabels()`'s fetch loses to the immediate WebSocket `running` tick. Both share the same root cause and the same fix; a live kiosk repro would confirm which. Either way, mode selection must not depend on a cache that is only populated by the selection UI event.

## Proposed fix

Derive countdown mode at run start from the **authoritative selected profile**, not the UI-populated cache:

- **Robust:** have `beginTimerMode()` fetch the estimate for the active profile (`activeRunProfileName` / server `selected_profile`) via `/profiles/details` at run start, instead of trusting `cachedEstimateSeconds`. This also covers runs started by the device's physical button, which bypass `notifyRun()`.
- **Localized (minimum):** `await loadProfileLabels(profile)` inside `notifyRun()` right before triggering the run (it already re-syncs `/profile/select` there), so the cache always matches the profile about to run.

## Acceptance criteria

- [ ] Run a profile with an estimate → countdown shows.
- [ ] After the run completes, hit Reset, run the same profile again → countdown **still** shows (not the stopwatch).
- [ ] A profile with no estimate still shows the stopwatch in both the first and subsequent runs.
- [ ] Add regression coverage for the reset → re-run path (the existing `tests/e2e/test_countdown_timer.py` only checks DOM presence, not a live run).

## Related

- Feature: #110 (issue) / #121 (PR)
- Interacting change: #275 (Reset clears `selected_profile` server-side)

jackhu23:
## What Changed

Fixes the countdown timer regressing to the stopwatch. The countdown-vs-stopwatch mode is now decided **at run start** from the authoritative active profile, instead of a client-side cache that only the profile-selection UI ever fills.

Closes #312.

## Why

A run with an estimate showed the **countdown** on the first run but flipped to the **stopwatch** on a subsequent run (reported as run → Reset → re-run). Diagnosed live: the deterministic trigger is the Run screen loading **while a run is active** (page reload / kiosk navigation / WS reconnect). `beginTimerMode()` fired on the first `running` tick and read a cold `cachedEstimateSeconds` (only warmed by a UI selection), locking the stopwatch for the whole run.

## Root cause (two layers)

1. **Client** — mode was read from `cachedEstimateSeconds`, never re-derived at run start.
2. **Server** — the intended fix re-fetches the estimate via `/profiles/details?name=<display name>`, but that endpoint's `?name=` lookup used a non-recursive `glob("*.json")` and **404'd every profile in `bundled/` or `local/`** (e.g. *Beer Spoilers - Bacteria*). So the client fetch fell through to its `.catch()` and the timer stayed on the stopwatch — the fix looked implemented but did nothing.

## The fix

- **`script.js`** — `beginTimerMode()` re-derives the mode at run start from `activeRunProfileName` (from the WS panel) via `/profiles/details?name=…`, with a monotonic token so a late fetch after Reset / a newer run is ignored. Selection cache kept only as a no-flash fast path. Also covers physical-button runs that bypass `notifyRun()`.
- **`main.py`** — `/profiles/details` `?name=` lookup: `glob` → **`rglob`**, so bundled/local profiles resolve (consistent with `_resolve_profile_display_name`, which produces the searched name).
- **`run.html`** — bump `script.js?v=68 → v=69` so devices load the new JS.

## Verification

Live, on a server carrying the fix vs. the old one — the exact call the timer makes:

| Server | `/profiles/details?name=Beer Spoilers - Bacteria` |
|---|---|
| **Fixed** | `estimated_completion_seconds=4080, time_unavailable=false` ✅ |
| Old | `404 / null` ❌ |

## Tests

- **Contract (new):** `test_profile_details_by_name_resolves_profile_in_subdir` — resolves a profile in a subdir by name. RED (`404`) before the `rglob` change, GREEN after. This is the real guard; the e2e tests **stub** `/profiles/details`, so they could not catch the 404.
- **E2E (new):** countdown shown on a cold cache, stopwatch stays stopwatch with no estimate, and the late-fetch-after-Reset race guard.
- Full contract + unit run: new test passes; 5 unrelated failures (bundled-profile-presence and device-cert tests) pre-exist on the pristine tree — verified by stash-and-rerun.

## Changes Made

- [x] Source code modified
- [x] Spec added (`specs/frontend/spec_countdown_reset_rerun.md`)
- [x] New tests written
- [x] Existing tests still pass (pre-existing failures unrelated)

## Hardware Testing

- [x] Tested in simulation mode only — timer/estimate behavior is frontend + HTTP; no hardware path involved.

## Reviewer Notes

The subtle bit: the e2e tests stub `/profiles/details`, which is why the earlier client-only fix passed CI but failed on the device. The contract test closes that gap. The race guard (`timerModeToken`) is what makes a mid-run reload safe — a late estimate response after Reset is dropped rather than flipping the timer back.

### DIFF
--- a/aquila_web/main.py
+++ b/aquila_web/main.py
@@ -1829,7 +1829,11 @@ async def profile_details(id: str | None = Query(default=None), name: str | None
         if not profile_path.exists():
             raise HTTPException(status_code=404, detail="Profile not found")
     else:
-        for path in profile_dir.glob("*.json"):
+        # Recurse into subdirs (bundled/, local/) — profiles do not live at the
+        # root. Mirrors _resolve_profile_display_name, which produces the very
+        # name searched here; a non-recursive glob would 404 every bundled/local
+        # profile looked up by name (#312).
+        for path in profile_dir.rglob("*.json"):
             try:
                 with path.open() as f:
                     data = json.load(f)
--- a/aquila_web/static/run.html
+++ b/aquila_web/static/run.html
@@ -164,7 +164,7 @@
     </main>
   </div>
   <script src="static/nav.js?v=1"></script>
-  <script src="static/script.js?v=68"></script>
+  <script src="static/script.js?v=69"></script>
   <script src="static/keyboard.js?v=14"></script>
 </body>
 </html>
--- a/aquila_web/static/script.js
+++ b/aquila_web/static/script.js
@@ -220,18 +220,56 @@ function setTimerLabel(text) {
   }
 }
 
-// Decide countdown vs. stopwatch for the active run and update the timer label.
-// Called when the running screen is entered; locks in the selected profile's estimate.
-function beginTimerMode() {
+// Monotonic token identifying the current run's timer session. Bumped on every
+// run-start and every reset so an estimate fetch that resolves late (after Reset
+// or a subsequent run) can detect it is stale and abstain (#312 race guard).
+let timerModeToken = 0;
+
+// Apply countdown-vs-stopwatch from an estimate value (null / non-positive =>
+// stopwatch) and update the timer label.
+function applyTimerMode(estimateSeconds) {
   countdownSeconds =
-    typeof cachedEstimateSeconds === "number" && cachedEstimateSeconds > 0
-      ? cachedEstimateSeconds
-      : null;
+    typeof estimateSeconds === "number" && estimateSeconds > 0 ? estimateSeconds : null;
   setTimerLabel(countdownSeconds != null ? TIMER_LABEL_REMAINING : TIMER_LABEL_ELAPSED);
 }
 
-// Reset timer to default stopwatch mode (run ended / stopped).
+// Decide countdown vs. stopwatch for the active run and update the timer label.
+// Called when the running screen is entered.
+//
+// The selection-warmed cache (`cachedEstimateSeconds`) is only a fast path: it is
+// populated solely by loadProfileLabels() on a UI selection, so after a run
+// completes and Reset clears the selected profile server-side (#275), a re-run of
+// the same profile re-enters `running` with a cold/stale cache and would wrongly
+// fall back to the stopwatch (#312). So we re-derive the mode from the
+// authoritative active profile (activeRunProfileName, set from the server panel)
+// via /profiles/details. This also covers runs started by the device's physical
+// button, which bypass notifyRun().
+function beginTimerMode() {
+  const token = ++timerModeToken;
+  applyTimerMode(cachedEstimateSeconds); // fast path: no flash when the cache is warm
+  const profileRef = activeRunProfileName;
+  if (!profileRef) {
+    return;
+  }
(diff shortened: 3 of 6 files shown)


NicoleCornell8 (approved):
Approving. Well-diagnosed two-layer fix (client cache re-derivation + server `rglob`). Race guard via `timerModeToken` correctly drops stale run-1 fetches across reset→rerun, and the contract test is the right guard for the `glob`→`rglob` 404 (e2e stubs `/profiles/details` so it couldn't catch it). A couple of low-severity notes (rglob name-collision is arbitrary first-match; fast-path retains a stale-warm cache on fetch error) — non-blocking.

Issue #354, opened 2026-07-28
I7: make the journal persistent so host-level history survives a reboot


nana-acorn:
**Priority: P3**

## Problem

`/etc/systemd/journald.conf` is at its default, so the journal is **volatile** — it lives in `/run/log/journal` (tmpfs) and is destroyed on every reboot.

That means host-level history is unrecoverable after a restart: `docker pull` events, container swaps, service restarts, kernel messages. During the last investigation, events from before Jul 20 were simply gone.

Disk is not a constraint (~42 GB free).

## Fix

```ini
[Journal]
Storage=persistent
SystemMaxUse=500M
MaxRetentionSec=90d
Compress=yes
```

`SystemMaxUse` protects the disk; `MaxRetentionSec` is the thing actually being bought. Both are enforced, whichever binds first — journald evicts oldest-first automatically, at file granularity, with no maintenance job.

Sizing should be confirmed against real volume on a device (`journalctl --disk-usage`) rather than guessed. The pairing matters because a chatty failure — a crash-looping container, Watchtower retrying a failed pull — can burn a size-only budget in days and evict exactly the history an incident makes valuable.

## This does not affect the API container

`fleet-config/docker-compose.yml:52-56` pins container logging separately:

```yaml
logging:
  driver: json-file
  options: { max-size: "10m", max-file: "3" }
```

Container logs go to Docker's json-file driver under `/var/lib/docker/containers/` — already on disk, already persistent, already capped at 30 MB per container. journald never sees them. Making the journal persistent changes nothing about the containers' I/O path.

## SD card wear is not a blocker

Devices boot from SD cards, so sustained writes are a fair concern — but journald is not where they come from:

| Source | Writes |
|---|---|
| **One OTA update** | **~889 MB** |
| journald at a 500 MB cap | 500 MB **total ceiling**, not per event |
| Container json-file logs | 30 MB ring per container |
| Optics log per run | ~1 MB |

A single update writes more to the card than the entire journal will ever hold. If SD longevity is a real concern, **#348 is the lever** — build caching takes updates from ~889 MB to ~4 MB, a ~200× cut in the single largest write source on the device. That is worth adding to #348's rationale, which currently argues size and speed only.

`SyncIntervalSec=300` can be added to batch flushes into fewer, larger writes if extra caution is wanted.

## ⚠️ Alloy's state is unverified — and it changes how much this matters

[...]

Issue #353, opened 2026-07-28
I6: Device clock: outbox Events are stamped with an unsynced device clock, corrupting warehouse timestamps


nana-acorn:
## Problem

A Raspberry Pi has no battery-backed RTC. On every power loss the clock resets and stays wrong until NTP resyncs.

That is usually filed as a logging annoyance. Here it is a **data integrity** problem, because the device's clock is what timestamps analytics data.

`aquila_web/local_db.py:12`:

```python
return datetime.utcnow().isoformat(timespec="seconds") + "Z"
```

Every Event written to the sync outbox is stamped with the **device's local clock**. So after a power loss:

1. Device boots with a bogus clock
2. Runs execute; Events are queued with bogus timestamps
3. NTP eventually resyncs
4. The outbox flushes — and those wrong timestamps are pushed to the analytics warehouse

The result is permanently corrupted timestamps in the warehouse, not just a messy local log. And the exposure window is the worst possible one: immediately after boot, when startup and first-run Events are being written.

There is **nothing in the repo** configuring `fake-hwclock`, `systemd-timesyncd`, `chrony`, or NTP — whatever the base Pi OS image does is what is running.

## Verify before fixing

Three questions decide the size of this:

1. **Does ingest trust the device timestamp, or stamp arrival time?** If `acorn-analytics` stamps on arrival, the damage is bounded to on-device logs and this drops in priority. If it trusts the device, warehouse data is already affected and this is urgent. **Check this first — it determines everything else.**
2. **Is `fake-hwclock` already installed?** Raspberry Pi OS ships it in most releases. If present, the question is whether it is *working*, not whether to install it.
3. **Is NTP actually syncing?** `timedatectl status` on a device. Given #314 was a DNS blackhole that silently stopped sync and OTA on the same devices, NTP failing quietly on those units is plausible.

## Fix

**a. Confirm/install `fake-hwclock`.** It saves the last known time to disk on shutdown and restores it at boot, before NTP syncs. Cheap, and it eliminates the "clock is at epoch zero" window.

**b. Confirm `systemd-timesyncd` or `chrony` is reaching an NTP server**, and that it is not silently failing the way DNS did in #314.

**c. Consider refusing to stamp Events until the clock is known good.** `systemd-time-sync.target` / `timedatectl` exposes whether time has been synchronised. Events written before that could be held, or marked with a `clock_unsynced` flag so the warehouse can identify suspect rows rather than silently trusting them.

[...]

Issue #351, opened 2026-07-28
I4: bake build identity into each image so a stale or mismatched version is detectable


nana-acorn:
**Priority: P1**

## Problem

There is no reliable way to know what software a Sentri is actually running.

Today the only identity signals are:

- `config_files/version.json` — `{"app_version": "2.1.0.5"}`, hand-bumped, says nothing about which build it came from
- `RUNNING_IMAGE_DIGEST` in `/opt/fleet/.env` — injected as an env var
- `GET /health` — returns `{"status": "ok"}` and nothing else

`RUNNING_IMAGE_DIGEST` is **not trustworthy**. `/update/apply` writes it into `/opt/fleet/.env` *before* triggering Watchtower (`main.py:2197-2215`), so it reflects what we *intended* to run, not what is running. If the pull fails, the recorded digest is wrong and nothing detects it (see #350).

Two specific failures follow, neither currently detectable:

- **A — API/UI version skew.** The two images can end up on different builds. #350 mode 5 describes one route in: if the API digest resolves and the UI digest fetch blips, an update is offered and applied with only one digest updated.
- **B — recorded vs. actual drift.** `.env` claims one digest; the container is running another. Nothing compares them, and the value we'd compare against is the untrustworthy one.

## Fix: bake build identity into the image

Write the build's git SHA and timestamp into each image **at build time**, and have each container report its own baked identity.

Because it is baked into the image bytes, it **physically cannot misreport** — unlike an injected env var, which is a claim made by whoever started the container.

This property is already safe to rely on: `config_files/` is **not** a mounted volume (`fleet-config/docker-compose.yml` mounts only `results`, `logs`, `profiles`, `config`, `data`), so it cannot be overwritten at runtime.

### Implementation

**CI** — `.github/workflows/docker-build.yml`, both build steps:

```yaml
build-args: |
  GIT_SHA=${{ github.sha }}
  BUILD_TIME=${{ github.event.head_commit.timestamp }}
```

**API image** — `docker/Dockerfile.api`, after `COPY . .`:

```dockerfile
ARG GIT_SHA=unknown
ARG BUILD_TIME=unknown
RUN python -c "import json,pathlib,os; \
p=pathlib.Path('config_files/version.json'); d=json.loads(p.read_text()); \
d['git_sha']=os.environ['GIT_SHA']; d['build_time']=os.environ['BUILD_TIME']; \
p.write_text(json.dumps(d, indent=4))"
```

[...]

Issue #350, opened 2026-07-28
I3: OTA status is unreliable in both directions: slow updates report failure, failed updates report success


nana-acorn:
## Summary

OTA update status is unreliable **in both directions**:

- A slow-but-successful update reports **"Update failed"**
- A failed check reports **"✓ Software is up to date"**
- A failed pull reports **"✓ Update complete — The device is running the latest software"**

The root cause is single: **we never observe the outcome, we infer it.** Watchtower already returns a report saying exactly what happened, and `main.py:2218` throws it away.

An operator cannot distinguish a healthy device from one silently stuck on old software. On a fleet that must stay patched, that is worse than a visible error.

---

## How the update works

**Phase 1 — check.** `_background_update_poller` runs every 300s → `_do_check_update` fetches manifest digests from GHCR over HTTPS. Never pulls.

**Phase 2 — apply.** `POST /update/apply` → backend calls Watchtower `POST /v1/update` → Watchtower pulls **~889 MB** and swaps containers.

Phase 2 is a multi-minute window on site wifi. Every failure mode below lives in one of these two phases.

---

## Why a network blip breaks the pull

A `docker pull` is not one download — it is a sequence of layer blobs streamed over separate HTTPS requests. For our API image that is ~889 MB across the base, apt, and pip layers, with the pip layer alone at ~419 MB.

When wifi drops mid-transfer:

1. The TCP connection carrying the in-flight layer breaks.
2. Docker retries that layer a limited number of times. If the network is still down, or comes back and drops again, the retries exhaust.
3. **The pull aborts.** Layers already fully downloaded are kept in the local store, but the partially-transferred layer is discarded — there is no resume from the byte offset on a later attempt.
4. Watchtower sees the pull error and **does not swap the container.** The old one keeps running.

Step 4 is correct behaviour — refusing to restart into an image you failed to fetch is exactly right. The bug is entirely on our side: by the time the pull fails, the backend has already reported the update as underway (or, per mode 1, already reported it as failed while it was still succeeding), and nothing ever goes back to check.

**Three things make this likely rather than theoretical:**

- **Window size.** 889 MB over site wifi is minutes of continuous transfer. Every second is exposure. The pull is that large because there is no CI layer cache (#348) — with caching, most updates would transfer ~4 MB, shrinking the window ~200×. **#348 is a partial mitigation for this issue.**
[...]

PR #367, opened 2026-07-29
fix(clock): make an untrustworthy device clock visible

nana-acorn:
Implements #353.

## Why this is a data problem, not a logging one

A Pi has no battery-backed RTC — after a power loss the clock resets and stays wrong until NTP resyncs. That is usually just messy logs. Here it corrupts the warehouse, because `aquila_web/local_db.py:12` stamps **every outbox Event** with the device's local clock:

```python
return datetime.utcnow().isoformat(timespec="seconds") + "Z"
```

Boot with a wrong clock → Events queue with wrong timestamps → the outbox flushes → those timestamps land permanently in the analytics warehouse. And the exposure window is the worst possible one: immediately after boot, while startup and first-run Events are being written.

Nothing in the repo configured `fake-hwclock`, `systemd-timesyncd`, `chrony` or NTP. Whatever the base Pi OS image does is what runs.

## Narrowing the window

New **Phase 3d** in `deployment2.sh`: install and enable `fake-hwclock` (restores the last known time at boot, before NTP syncs) and enable NTP + `systemd-timesyncd`.

## Making it visible

Neither of those helps a device that **never reaches an NTP server** — which, given #314 was a DNS blackhole on these same devices, is not hypothetical.

`aquila_web/clock_health.py` combines two independent signals:

**1. systemd's NTP sync flag.** Authoritative, but only visible inside the container if `/run/systemd/timesync` is bind-mounted — added to fleet-config, read-only.

**2. A lower bound from the image's own baked `build_time` (#351).** A device cannot be running software that has not been built yet, so a clock earlier than the build is **definitely** wrong. Needs no mount, and catches the dominant failure: an RTC-less Pi booting to the epoch.

### `unknown` is a distinct verdict, on purpose

Signal 2 is **one-sided by design** — it proves a clock is wrong, never that it is right. A clock set to 2099 passes it.

So `clock_trusted` is true **only** for `synced`. A clock that cannot be verified reports `unknown`, not trusted. There is a test documenting the far-future limitation explicitly, so it reads as a known boundary rather than an oversight.

```
GET /clock
{ "clock_status": "unknown",
  "clock_trusted": false,
  "clock_reason": "NTP sync state unavailable and no baked build time to compare against",
  "ntp_synchronized": null,
  "clock_before_build": null }
```

A new endpoint rather than an addition to `/health`, to avoid conflicting with the `/health` rewrite in #351.

## Verification checks assert the clock is *set*

`deployment2_verify.sh` gains a Phase 3d that checks `NTPSynchronized`, **not** merely that `timesyncd` is active. A device that cannot reach an NTP server has the service running and a wrong clock — the same "verified the wrong thing, concluded success" pattern as #350 and the Alloy checks in #354.

## Independent of #351

`build_time` is read defensively; its absence degrades to `unknown` rather than failing. Verified in a container with neither signal available: reports unknown/untrusted and says why. Once #351 lands the lower-bound check activates automatically.

## One guardrail test extended

`test_linux_compatibility` asserts device volumes come from `/opt/aquila`, `/opt/fleet` or `/root`. It caught the new mount, which is what it is for. Added to its explicit exemption list alongside the Docker socket, with the reason.

## ⚠️ Not verified, and one open question

**No hardware verification.** sn03 went offline mid-work, so `fake-hwclock`'s presence, the NTP state, and any actual drift are all unchecked on a real device.

**The question that decides this issue's true priority is still open:** does `acorn-analytics` ingest trust the device timestamp, or stamp arrival time? If it stamps on arrival, the damage is bounded to on-device logs and this drops down the list. If it trusts the device, warehouse data is already affected. That check needs someone with access to the ingest pipeline.

Worth running against the warehouse either way: query for Events with implausible timestamps — epoch-adjacent, or out of order within a single run — to size any existing damage.

## Tests

13 unit tests covering the lower-bound check, the sync flag (including that an unmounted directory is `unknown`, not `unsynced`), the combined verdict precedence, and graceful degradation.

**Full suite:** 37 failed / 855 passed / 237 skipped — the 37 match clean `main`, verified by stashing and diffing the failure lists.

Part of #353.

### DIFF
--- a/aquila_web/clock_health.py
+++ b/aquila_web/clock_health.py
@@ -1,0 +1,120 @@
+"""Is this device's clock trustworthy? (#353)
+
+A Raspberry Pi has no battery-backed RTC. After a power loss the clock resets
+and stays wrong until NTP resyncs — and `aquila_web/local_db.py:12` stamps every
+outbox Event with `datetime.utcnow()`. So Events queued in that window carry
+wrong timestamps and are flushed to the analytics warehouse that way. The
+exposure window is the worst one: immediately after boot, while startup and
+first-run Events are being written.
+
+Two independent signals, because neither alone is sufficient:
+
+1. systemd's NTP sync flag. Authoritative when available, but the backend runs
+   in a container and only sees it if /run/systemd/timesync is bind-mounted.
+
+2. A lower bound from the image's own baked build_time (#351). A device cannot
+   legitimately be running software that has not been built yet, so a clock
+   earlier than the build is definitely wrong. This needs no mounts and works
+   everywhere, and it catches the dominant failure — an RTC-less Pi booting to
+   the epoch or to a stale saved time.
+
+Signal 2 is a one-sided test: it proves a clock is wrong, never that it is
+right. A clock set far into the future passes it. That is why the sync flag is
+preferred when present, and why `unknown` is a distinct answer from `synced`.
+"""
+
+from __future__ import annotations
+
+import os
+from datetime import datetime, timezone
+from pathlib import Path
+
+# Written by systemd-timesyncd once it has successfully synchronised.
+SYNC_FLAG_PATH = os.getenv("AQ_TIMESYNC_FLAG", "/run/systemd/timesync/synchronized")
+
+# Status values
+SYNCED = "synced"
+UNSYNCED = "unsynced"
+UNKNOWN = "unknown"
+
+
+def _parse_iso(value: str | None) -> datetime | None:
+    if not value or value == "unknown":
+        return None
+    try:
+        parsed = datetime.fromisoformat(str(value).replace("Z", "+00:00"))
+    except (ValueError, TypeError):
+        return None
+    if parsed.tzinfo is None:
+        parsed = parsed.replace(tzinfo=timezone.utc)
+    return parsed
+
+
+def clock_is_before_build(build_time: str | None, now: datetime | None = None) -> bool | None:
+    """True when the clock predates the running image's build — definitely wrong.
+
+    None when it cannot be evaluated (no baked build_time, e.g. an image from
+    before #351). None is not False: an unevaluable check must not read as a pass.
+    """
+    built = _parse_iso(build_time)
+    if built is None:
+        return None
+    current = now or datetime.now(timezone.utc)
+    if current.tzinfo is None:
+        current = current.replace(tzinfo=timezone.utc)
+    return current < built
+
+
+def ntp_sync_flag() -> bool | None:
+    """systemd's view of whether time has been synchronised.
+
+    None when the flag is not visible — which is the normal case in a container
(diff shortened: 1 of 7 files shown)


nana-acorn:
## Correction: the fake-hwclock framing overstates what is being fixed

The PR reads as though `fake-hwclock` is missing. It probably is not — Raspberry Pi OS ships it by default in most releases, so `apt-get install -y fake-hwclock` is very likely a no-op.

It is harmless either way (apt is idempotent), but the emphasis is wrong. **This was never verified on a device** — sn03 went offline before the check could run.

### What is actually new here

- **NTP enablement and verification** — nothing in the repo configured `systemd-timesyncd`, NTP, or chrony before this
- **The `NTPSynchronized` check** rather than `is-active` — a device that cannot reach an NTP server has the service running and a wrong clock
- **`GET /clock`** — the clock state was not observable at all
- **The `build_time` lower bound** — a device cannot be running software that has not been built yet

None of those depend on whether `fake-hwclock` was already present.

### Still unverified, needs one device

```bash
dpkg -l fake-hwclock | tail -1          # installed?
systemctl is-enabled fake-hwclock       # enabled?
cat /etc/fake-hwclock.data              # current, or stale?
timedatectl status                      # is NTP actually synchronized?
date -u                                 # how far off is it right now?
```

If `fake-hwclock` turns out to be installed and working everywhere, the install line stays as a no-op safeguard for a device where it is not, and the commit message should be corrected to say so.

### And the question that decides this issue's priority is still open

Does `acorn-analytics` ingest trust the device timestamp, or stamp arrival time? If it stamps on arrival, the damage is bounded to on-device logs and this drops down the list. That needs someone with ingest-pipeline access, not a device.

nana-acorn:
## Delivery: this won't reach existing devices

The clock/timesync host config here lands in `deployment2.sh`, which runs **once, at provisioning**. Nothing re-runs it, so this reaches newly built devices only — every Sentri already in the field keeps its current (unconfigured) clock setup indefinitely.

That's the same gap behind #357: `/opt/fleet/update.sh` is also written once by `deployment2.sh`, which is why an image pull with no retry at all survived on every device for months.

It keeps landing there because nothing else has the privilege — containers can't write host system config, and `deployment3_greengrass.sh` provisions the nucleus with `--component-default-user ggc_user:ggc_group`, so component lifecycles can't either.

**#397 now proposes the mechanism**: a root-run Greengrass maintenance component that idempotently re-asserts host config on every deployment. Suggest re-pointing the `deployment2.sh` portion of this PR at that component once it exists, so the fix actually reaches the fleet.

The application half of this PR (`clock_health.py`, `/health` reporting) is unaffected — it ships in the image and reaches devices normally. Only the host-config portion has the delivery problem.

Worth deciding before merge: land as-is and re-point later, or hold the host-config part until the component exists.

NicoleCornell8 (changes requested):
this should be for greengrass branch

Issue #348, opened 2026-07-28
I1: Shrink and speed up the API container: no CI layer cache means every OTA ships 889 MB


nana-acorn:
## Problem

Every OTA update ships the **entire ~889 MB image** to every Sentri in the fleet, even for a one-line code change. CI takes a flat ~7 minutes per build regardless of what changed.

Both symptoms have the same root cause: **there is no Docker layer cache in CI**, so the 419 MB `pip install` layer is rebuilt with a fresh digest on every push. Watchtower dedupes on layer digest, so a new digest means a full re-pull fleet-wide.

## Current baseline (measured)

Built from a clean `git archive HEAD` export on `linux/arm64` to match CI.

### Image layers

| Layer | Size |
|---|---|
| `apt-get install build-essential procps sudo curl` | **292 MB** |
| `pip install -r requirements-backend.txt` | **419 MB** |
| `python:3.11-slim-bookworm` base | ~174 MB |
| `COPY . .` (app code) | 3.9 MB |
| **Sum of layers** | **889 MB** |
| `docker images` reports | 1.15 GB |

Build context is only 7.4 MB — essentially all of the image is base OS + apt + pip.

### `site-packages` breakdown (406 MB total)

| Package | MB | Verdict |
|---|---|---|
| **pandas** | **79** | **REMOVE — redundant** |
| numpy | 71 | keep (16 files) |
| matplotlib | 38 | keep — required |
| twisted | 36 | forced by `simple-rpc` |
| fontTools | 29 | matplotlib dep |
| pillow.libs + PIL | 25 | matplotlib dep |
| pytest + pygments | 15 | forced by `simple-rpc` (upstream packaging bug) |
| autobahn | 10 | forced by `simple-rpc` |

### CI build durations (last 8 runs)

`6.8, 7.1, 7.1, 7.2, 7.4, 7.5, 7.5, 7.7` minutes.

The flatness is the tell — a code-only change costs exactly as much as a dependency bump, because nothing is cached.

## Target

| | Now | After |
|---|---|---|
| Image (sum of layers) | 889 MB | **~540 MB** |
| Bytes pulled per device, code-only change | 889 MB | **~4 MB** |
| CI time, code-only change | ~7 min | **< 1 min** |

---

# Work items, in priority order

## 1. Add `cache-from` / `cache-to` to CI — 889 MB → ~4 MB per device pull

**Highest impact, smallest change.** Do this first.

[...]

Issue #357, opened 2026-07-28
OTA: retry the image pull, and force IPv6 as a last resort when IPv4 can't carry it


nana-acorn:
Branch: `fix/ota-pull-retry-ipv6-fallback` (3 commits)

## Problem

On some sites the OTA update fails at the **image transfer** step, not the check step.
The device sees the new digest on GHCR (`/update/status` reports `available`), the
operator presses Update, the layer download starts — and it stalls or errors out.

**Field evidence:** on an affected device (Pigmind), the operator SSH'd in, switched
IPv4 off so the device had only IPv6, and the update completed immediately. That
observation is what this issue is built on.

## Two separate problems, one of which is bigger

Investigating turned up a plain bug that has nothing to do with address families:

**1. The pull had no retry at all.** `scripts/deploy/fleet-update.sh` ran a bare
`docker compose pull` under `set -euo pipefail`. A single dropped connection aborted
the whole update *before* the digest write-back, leaving the device on the old image.
Every site is exposed to this, including the ones with no IPv6. This is the higher
priority fix and it stands on its own.

**2. Some sites can't carry the transfer over IPv4.** The IPv6 rescue below. Narrower,
and — see the caveat — only applicable in specific network conditions.

## Why IPv6 helps, and why only sometimes

A pull is two conversations with two different GitHub hosts:

| Leg | Host | Size | Has AAAA? |
|---|---|---|---|
| Auth + manifest | `ghcr.io` | a few KB | **no** |
| Blob / layer download | `pkg-containers.githubusercontent.com` | hundreds of MB | **yes** |

(Checked against 8.8.8.8, 1.1.1.1 and 9.9.9.9 — `ghcr.io` publishes no AAAA record.)

This matches the symptom: the small manifest fetch succeeds over a degraded IPv4 path,
so the device correctly reports an update available, while the bulk transfer dies.

**What does not work:** merely *enabling* IPv6 and trusting address selection to prefer
it. Go (dockerd) and glibc already prefer IPv6 over IPv4 wherever both work, so a device
with usable IPv6 is already trying it — making it "usable" changes nothing. An earlier
revision of this branch did that; it was a no-op and has been removed.

**What does work** is what was done by hand in the field: take IPv4 away entirely so the
pull has no choice. Because `ghcr.io` has no AAAA, that can only complete on a network
running **NAT64/DNS64** — an IPv6-mostly setup that synthesises addresses for IPv4-only
[...]

PR #368, opened 2026-07-29
chore(observability): persist the journal and verify logs actually ship

nana-acorn:
Implements #354.

## Persistent journal

The journal defaults to volatile storage (`/run/log/journal`, a tmpfs), so **every reboot destroys host-level history** — docker pull events, container swaps, service restarts, kernel messages. During the last investigation everything before Jul 20 was simply gone.

New **Phase 3e** in `deployment2.sh` installs a journald drop-in:

```ini
Storage=persistent      # survive reboots
SystemMaxUse=500M       # protect the SD card
MaxRetentionSec=90d     # the window actually being bought
Compress=yes            # stated so a future edit doesn't silently drop it
SyncIntervalSec=300     # fewer, larger writes — the pattern SD cards prefer
```

**Both caps, deliberately.** A size cap alone lets a chatty failure — a crash-looping container, a retrying pull — burn the budget in days and evict precisely the history an incident makes valuable.

**Container logs are untouched.** `fleet-config` already pins those to Docker's json-file driver at 10m × 3, and journald never sees them. A test asserts that stays true, since conflating the two would silently uncap container logging.

**SD wear is not a blocker.** One OTA writes ~889 MB (dropping to ~4 MB once #348's cache lands) against a journal ceiling of 500 MB *total*.

## Alloy checks now assert delivery, not liveness

The existing three checks:

```bash
check 13 "alloy installed"       "which alloy"
check 13 "alloy service active"  "systemctl is-active alloy.service | grep -q active"
check 13 "alloy service enabled" "systemctl is-enabled alloy.service | grep -q enabled"
```

**All three pass while Alloy ships nothing at all** — bad Grafana Cloud credentials, an unreachable endpoint, a misconfigured `loki.write`. The process is up, `systemctl` is happy, the deploy verifies clean, and no log line ever arrives.

Same "verified the wrong thing, concluded success" pattern as #350. If the pipeline is in fact dead, this is very likely **why nobody noticed**.

Added checks read Alloy's own metrics for entries sent and entries dropped.

## `max_age` raised from 24h to 168h

`loki.source.journal` only reads back `max_age`, so anything older is **never shipped** even with a persistent journal — silently capping what Grafana Cloud can hold after an outage.

These devices are routinely offline for days; sn02 was last seen 53 days ago. Bounded by the journal's own 90d retention rather than left unlimited, with a test asserting that relationship so the two cannot drift apart.

## Retrieval

Deliberately **no** diagnostic-bundle tooling. An earlier draft proposed one on the reasoning that "persisted ≠ reachable" — that was wrong. Tailscale is on every device, so once one is online `journalctl` reads whatever the retention window holds.

In fact SSH + persistent journal is *better* than Alloy for the case that matters: Alloy backfills at most `max_age`, while `journalctl` reads the full 90 days. At 8 devices, per-device SSH is not a constraint worth building around.

## ⚠️ Not verified on hardware

No device was reachable. Two things need a live one:

**1. Is Alloy delivering at all?** One Loki query settles it — and determines how much of this issue was even needed. If it works, persistent journal is a backfill buffer; if it doesn't, the local journal is the only record.

**2. The exact metric names.** `loki_write_sent_entries_total` and `loki_write_dropped_entries_total` are the expected names but were **not confirmed against a running Alloy**. Run `curl -s localhost:12345/metrics | grep loki_write` on a device before merging — a check that silently never matches is worse than no check. The default listen address `127.0.0.1:12345` also needs confirming for this install.

## Tests

9 config assertions: journal persistence and both caps, container logging still pinned separately, Alloy checks asserting delivery, `max_age` above 24h and within journal retention, and that the new phase is registered in `ALL_PHASES` (a phase not listed there never runs).

**Full suite:** 37 failed / 851 passed / 237 skipped — the 37 match clean `main`. All touched scripts pass `bash -n`.

Part of #354.

### DIFF
--- a/aquila_web/main.py
+++ b/aquila_web/main.py
@@ -22,6 +22,50 @@ from aq_curve.analysis_service import AnalysisService
 logger = logging.getLogger( __name__ )
 logger.setLevel("WARNING")
 
+
+# ── Quiet the polling endpoints in the access log (#354) ──────────────────────
+#
+# Uvicorn logs every HTTP request. The kiosk polls a handful of endpoints
+# continuously, so the access log is overwhelmingly `GET /button_status/ 200 OK`.
+# Measured on sn08: aquila-backend produced 19 MB of container log in two days —
+# 48x the next container, and ~9.5 MB/day of almost entirely poll noise. That
+# buries the lines that matter (update decisions, sync results, errors), and once
+# container logs are shipped to Grafana Cloud it would be paid for twice.
+#
+# Only SUCCESSFUL polls are dropped. A failing poll still logs, so the endpoint
+# going wrong stays visible — which is the whole reason to keep an access log.
+_QUIET_ACCESS_PATHS = frozenset({
+    "/button_status", "/button_status/",  # the kiosk's main poll loop
+    "/health",                            # Docker HEALTHCHECK, every 30s
+    "/results", "/results/status",
+    "/run/name",
+    "/drawer/state",
+    "/update/status",
+})
+
+
+class _QuietPollFilter(logging.Filter):
+    """Drop access-log records for successful polls of high-frequency endpoints."""
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        args = getattr(record, "args", None)
+        # uvicorn.access passes (client_addr, method, full_path, http_version, status)
+        if not isinstance(args, tuple) or len(args) < 5:
+            return True  # unrecognised shape — keep it rather than lose a line
+        path, status = args[2], args[4]
+        if not isinstance(path, str):
+            return True
+        try:
+            failed = int(status) >= 400
+        except (TypeError, ValueError):
+            return True
+        if failed:
+            return True
+        return path.split("?")[0] not in _QUIET_ACCESS_PATHS
+
+
+logging.getLogger("uvicorn.access").addFilter(_QuietPollFilter())
+
 app = FastAPI(redirect_slashes=False)
 static_dir = Path(__file__).parent / "static"
 
--- a/scripts/deploy/deployment2.sh
+++ b/scripts/deploy/deployment2.sh
@@ -177,6 +177,57 @@ run_test "NM dispatcher installed" "test -x /etc/NetworkManager/dispatcher.d/99-
 phase_pass "NetworkManager BSSID dispatcher installed"
 
 # ═══════════════════════════════════════════════════════════════════════════════
+# Phase 3e — Persistent journal (#354)
+# ═══════════════════════════════════════════════════════════════════════════════
+phase_start "3e" "Persistent Journal"
+
+# The journal defaults to volatile storage (/run/log/journal, a tmpfs), so every
+# reboot destroys the host-level history: docker pull events, container swaps,
+# service restarts, kernel messages. During the last investigation everything
+# before Jul 20 was simply gone.
+#
+# This does NOT affect container logs. fleet-config pins those to Docker's
+# json-file driver (10m x 3), which already lands on disk under
+# /var/lib/docker/containers. journald never sees them.
+#
+# Devices boot from SD cards, so writes matter — but not much next to what is
+# already happening: one OTA writes ~889 MB (dropping to ~4 MB once the build
+# cache from #348 is in), against a journal ceiling of 500 MB TOTAL.
+mkdir -p /etc/systemd/journald.conf.d
+cat > /etc/systemd/journald.conf.d/10-aquila-persistent.conf <<'EOF'
+[Journal]
+# Survive reboots. Without this the journal lives in a tmpfs and is destroyed on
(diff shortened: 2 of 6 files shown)


nana-acorn:
## Revised, with measurements from three live devices

The premise of this issue was wrong in an important way. Checked sn01 (burner), sn05 (pilot) and sn08 (pilot, 76 assays over 18 days).

### Alloy is not broken — it was never configured

```
sn01   config.alloy 696 bytes, Jul 22    journal read: 90    sent: 176
sn05   config.alloy 696 bytes, Jul  1    journal read: 85    sent: 170
sn08   config.alloy 696 bytes, Jul 10
```

All three have only `remotecfg`, `loki.write` and `prometheus.remote_write` — **destinations with no sources**. The only components running are `self_monitoring_logs_linux` and `self_monitoring_metrics`: Alloy shipping its own logs about itself.

Delivery works — 170–176 entries, zero dropped, valid credentials. The device journal simply has nothing pointing at it.

**Cause:** `setup_grafana_alloy_rpi.sh` is never invoked by any deploy script. It was run by hand on sn01 on 18 March (its `.bak` proves it), and `deployment2.sh:1082` runs Grafana's `install-linux.sh` on every deployment, which writes a fresh config. Three devices, three different reset dates, identical outcome.

So the fix is smaller than this issue assumes: **call `setup_grafana_alloy_rpi.sh` from `deployment2.sh` after the installer.** Its existing `grep -q "integrations/raspberrypi-node"` guard makes that idempotent.

### The delivery checks are now verified against hardware

`loki_write_sent_entries_total` and port `12345` were flagged in this PR as unconfirmed guesses. Both are correct — confirmed on all three devices. And they catch exactly this failure: `systemctl is-active alloy` passes on every device while the journal goes unshipped.

### Log volume — measured, and it inverts an earlier assumption

sn08, per container over 2 days:

```
aquila-backend    19 MB     <- 48x the next
aquila-app       1.7 MB
aquila-ui        396 KB
aquila-watchtower 12 KB
                 ------
                  22 MB     ~= 11 MB/day
```

I previously said `aquila-ui` was the noisy one. Wrong — the **backend** dominates, because uvicorn logs every request and the kiosk polls continuously.

Addressed at the source in the latest commit: successful polls to `/button_status`, `/health`, `/results`, `/run/name`, `/drawer/state`, `/update/status` no longer log. Failing ones still do. Verified in a container — 30 polls produced 0 lines, a real request and a 404 produced 1 each.

### Sizing confirmed

```
container logs   ~11 MB/day (before the filter)
journal            8 MB total
7-day total      ~85 MB
```

`SystemMaxUse=500M` holds ~45 days at that rate, so the **7-day `max_age` window governs and the size cap never binds** — which is what the 7-day decision requires. Roughly 6x headroom, and more after the filter.

### Also confirmed: sn08's clock is correct

```
device: 2026-07-29T18:50:11Z
local:  2026-07-29T18:50:11Z    exact match
NTPSynchronized: yes
```

Relevant to #353 — no active drift on this device, which lowers that issue's urgency.

### Container logs are still not collected

Driver is `json-file` on all containers, so they are not in the journal and not shipped. Two routes: a `loki.source.file` for `/var/lib/docker/containers/*/*-json.log`, or switching the driver to `journald` (which keeps `docker logs` working — I was wrong to say otherwise) so the existing journal pipeline picks them up.

The second reuses a pipeline that already works and needs no Docker socket. Worth doing, but it is a separate change from this PR.

nana-acorn:
## Delivery: this won't reach existing devices

The journald drop-in (`/etc/systemd/journald.conf.d/10-aquila-persistent.conf`, `Storage=persistent`, `SystemMaxUse=500M`) is written by `deployment2.sh`, which runs **once, at provisioning**. Nothing re-runs it, so this reaches newly built devices only — every device already in the field keeps losing its journal at each reboot, which is precisely what #354 is about.

Worth noting the irony: the devices with the most lost history are the oldest ones, and they're exactly the ones this wouldn't reach.

Same gap behind #357 — `/opt/fleet/update.sh` is also written once by `deployment2.sh`, which is why an image pull with no retry at all survived on every device for months.

It keeps landing there because nothing else has the privilege: writing `/etc/systemd/journald.conf.d` and restarting `systemd-journald` needs root on the host. Containers can't, and `deployment3_greengrass.sh` provisions the nucleus with `--component-default-user ggc_user:ggc_group`, so component lifecycles can't either.

**#397 now proposes the mechanism**: a root-run Greengrass maintenance component that idempotently re-asserts host config on every deployment (write drop-in, reload only if changed). Suggest re-pointing the `deployment2.sh` portion at that component once it exists.

The application half (`test_access_log_filter.py`, the log filter in `main.py`) is unaffected — it ships in the image and reaches devices normally.

Worth deciding before merge: land as-is and re-point later, or hold the host-config part until the component exists.

PR #369, opened 2026-07-29
fix(ota): retry the image pull, force IPv6 as a last resort, and publish the host updater

### LINKED ISSUE #357: OTA: retry the image pull, and force IPv6 as a last resort when IPv4 can't carry it

nana-acorn:
Closes #357.

## What this fixes

**1. The image pull had no retry at all.** `fleet-update.sh` ran a bare `docker compose pull`
under `set -euo pipefail`, so one dropped connection aborted the update *before* the digest
write-back and left the device on the old image. This affects every site, including those with
no IPv6, and is the higher-priority fix here.

**2. Some sites can't carry the transfer over IPv4.** In the field, switching a stuck device
(Pigmind) to IPv6 by hand completed the update immediately.

**3. The host updater could not be updated.** `/opt/fleet/update.sh` was written once by
`deployment2.sh` at provisioning and never refreshed — so every device runs whatever `main`
said the day it was set up, and a fix to it could only reach the fleet by hand. That is *why*
problem 1 survived so long on every device.

## How

**Retry with classification** — transport failures (timeout, reset, unexpected EOF, TLS,
partial layer) retry; auth (401/403) and unknown tags (404) fail fast, because no retry or
address family turns a 401 into a 200 and making an operator wait out a doomed retry is worse
than telling them straight away. Budget: 8 minutes hard cap — 3 IPv4 attempts, then up to 2
IPv6 attempts, each bounded by the remaining budget.

**IPv6 as a last resort, forced rather than merely enabled.** An earlier revision made IPv6
"usable" and trusted address selection to prefer it; that was a no-op, since Go (dockerd) and
glibc already prefer IPv6 wherever both work. What worked in the field was removing IPv4
entirely, so that is what this does — dropping the IPv4 *default route* for one attempt.

Because `ghcr.io` publishes no AAAA record, a v6-only pull can only authenticate where the
network runs NAT64/DNS64. Reaching `ghcr.io` over IPv6 tests for exactly that and gates the
attempt, so sites that cannot benefit never lose IPv4 and fail after the IPv4 retries instead.

Safety for the forced attempt: only the default route is removed, never an address or
interface; restored by a trap on any exit, by a detached watchdog if the shell is killed
outright, and by a reboot regardless since nothing is persisted.

**Host updater published from the image.** The backend syncs its bundled `fleet-update.sh` to
`/opt/fleet/update.sh` on startup, using the `COPY . .` already in `Dockerfile.api` and the
`/opt/fleet` mount already present for the reboot sentinel. The updater now rides normal ring
promotion. Written via `os.replace` (atomic): a half-written `update.sh` would leave a device
unable to update at all, and the only tool that could repair that remotely is the script being
written.

**`/update/status` reports the outcome** via a `last_pull` breadcrumb — result, families tried,
detail — so a failed update is diagnosable without SSH. Under its own key rather than folded
into `error`, because it describes the scripted path only: the Update button drives Watchtower,
which performs its own pull and never runs this script.

## Tests

45 new. `tests/fleet_device/test_pull_retry.py` sources the script and executes its shell
functions rather than asserting on source text; `unit_tests/test_fleet_update_sync.py` covers
the publish; `tests/contract/test_update_pull_outcome.py` covers the API.

Cases worth calling out: a digest containing `401403404` must not be read as an auth failure;
a failed pull must never report success when `pipefail` is unset; the IPv4 route must be
restored on the failure path; a simulated disk-full mid-write must leave a working updater.

Suite: 55 pre-existing failures unchanged before and after, +45 passing.

## Rollout note — the first update does not benefit

A device still running the old script performs its next update the old way (no retry). That
update ships the new image, whose backend then publishes the new updater. **From the second
update onward the retry and fallback are live.**

The exception matters: a device *currently failing* to update cannot receive the fix this way,
since receiving it requires a working update. Those need one hand-push (or a network fix) first.

## Unverified

- The NAT64 explanation for Pigmind is inference, not measurement — that device was not
  reachable. If wrong, the gate simply never fires and the retry fix is unaffected.
- The forced-IPv6 attempt has not run against a real degraded network, only stubs.
- Watchtower still performs its own unretried pull; this covers the scripted path only.

### DIFF
--- a/aquila_web/main.py
+++ b/aquila_web/main.py
@@ -9,10 +9,12 @@ from fastapi.staticfiles import StaticFiles
 from pathlib import Path
 from fastapi import WebSocket
 import asyncio
+import contextlib
 import logging
 from datetime import datetime
 import os
 import sys
+import tempfile
 from pydantic import BaseModel
 from typing import Optional
 import json
@@ -2248,6 +2250,74 @@ async def approve_update_gate():
 
 _PULL_OUTCOME_PATH = os.getenv("AQ_PULL_OUTCOME_PATH", "/opt/fleet/last_pull.json")
 
+# --- keeping the host updater current ---------------------------------------
+# /opt/fleet/update.sh is the host-side updater. Until now it was written exactly once,
+# by deployment2.sh at provisioning, and nothing ever refreshed it -- so every device
+# ran whatever main said on the day it was set up, and a fix to it could only reach the
+# fleet by hand. That is why the un-retried pull (#357) survived so long.
+#
+# The script ships inside this image (Dockerfile.api does `COPY . .`) and /opt/fleet is
+# mounted read-write here, so syncing it on startup puts the updater on the normal ring
+# promotion path: sandbox -> dev -> pilot -> prod, like everything else. Devices that
+# are offline today pick it up whenever they next update.
+_FLEET_UPDATE_SRC = os.getenv(
+    "AQ_FLEET_UPDATE_SRC", "/opt/aquila/scripts/deploy/fleet-update.sh"
+)
+_FLEET_UPDATE_DEST = os.getenv("AQ_FLEET_UPDATE_DEST", "/opt/fleet/update.sh")
+
+
+def sync_fleet_update_script(src: str | None = None, dest: str | None = None) -> str:
+    """Publish this image's fleet-update.sh to the host. Returns what it did.
+
+    Written to a temporary file and moved into place with os.replace, which is atomic
+    on the same filesystem. This matters more than it looks: a half-written update.sh
+    would leave the device unable to update at all, and the only tool that could fix
+    that remotely is the very script being written.
+
+    The image is the source of truth, so a device hand-patched for debugging is
+    restored on the next restart -- deliberate, and the documented rollback already
+    relies on an image update healing such edits.
+
+    Never raises. A read-only mount or a missing source is degraded operation (the
+    device keeps its existing updater), not a reason to fail startup.
+    """
+    src = src or _FLEET_UPDATE_SRC
+    dest = dest or _FLEET_UPDATE_DEST
+
+    try:
+        with open(src, "rb") as f:
+            desired = f.read()
+    except OSError:
+        return "skipped"
+
+    try:
+        with open(dest, "rb") as f:
+            if f.read() == desired:
+                return "unchanged"   # avoid rewriting the file on every boot
+    except OSError:
+        pass   # absent or unreadable -> fall through and write it
+
+    try:
+        directory = os.path.dirname(dest) or "."
+        fd, tmp = tempfile.mkstemp(dir=directory, prefix=".update.sh.")
+        try:
+            with os.fdopen(fd, "wb") as f:
+                f.write(desired)
+                f.flush()
+                os.fsync(f.fileno())
+            os.chmod(tmp, 0o755)
+            os.replace(tmp, dest)
+        except BaseException:
+            with contextlib.suppress(OSError):
+                os.unlink(tmp)
+            raise
+    except OSError as e:
+        logger.warning("could not sync %s to %s: %s", src, dest, e)
+        return "skipped"
+
+    logger.info("synced host updater to %s", dest)
+    return "synced"
+
 
 def _read_last_pull() -> dict | None:
     """Outcome of the last device-side pull, written by scripts/deploy/fleet-update.sh.
@@ -2480,6 +2550,16 @@ async def resolve_update_completion() -> None:
         _resolve_startup_update_state()
     except Exception as e:  # noqa: BLE001 - never block startup on this
         logger.warning("update sentinel resolution failed: %s", e)
+
+
+@app.on_event("startup")
+async def publish_host_updater() -> None:
+    """Keep /opt/fleet/update.sh in step with the image, so fixes to the host updater
+    ride the normal ring promotion instead of needing a hand-push to every device."""
+    try:
+        sync_fleet_update_script()
+    except Exception as e:  # noqa: BLE001 - never block startup on this
+        logger.warning("host updater sync failed: %s", e)
 
 
 @app.on_event("startup")
--- a/aquila_web/main.py
+++ b/aquila_web/main.py
@@ -2246,6 +2246,32 @@ async def approve_update_gate():
     return update_gate.GATE.status()
 
 
+_PULL_OUTCOME_PATH = os.getenv("AQ_PULL_OUTCOME_PATH", "/opt/fleet/last_pull.json")
+
+
+def _read_last_pull() -> dict | None:
+    """Outcome of the last device-side pull, written by scripts/deploy/fleet-update.sh.
+
+    Surfaced so a failed update says which address families were tried — enough to
+    tell "no IPv6 at this site" from "registry unreachable" from "bad token" without
+    SSHing in (#357).
+
+    Scope caveat: the operator's Update button drives Watchtower, which performs its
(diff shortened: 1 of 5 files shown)


nana-acorn:
## Decision: rework for Greengrass rather than merge to `main`

The Greengrass migration lands soon, so this moves into the fleet-hardening epic and gets reworked against `feat/greengrass-migration` instead of merging here.

Worth recording what actually survives, because it is less than it looks:

**The retry loop probably does not port directly.** Under Greengrass the image transfer is performed by `aws.greengrass.DockerApplicationManager`, which pre-stages ECR images before the compose stack runs. `docker compose up -d` in the recipe's `Run` step finds them already local. So there is no `docker compose pull` of ours left to wrap in a retry — the pull belongs to the nucleus, which we do not control. Greengrass has its own deployment failure and rollback semantics in that role.

**The IPv6 fallback changes shape.** The underlying problem is real and field-confirmed (#357: a device whose IPv4 path could not carry the transfer completed its update once forced onto IPv6). But under Greengrass this stops being pull logic and becomes **host network configuration** — making the device's network capable of the transfer, rather than retrying a command we own. That belongs in the maintenance component proposed in #397, alongside the other host convergence.

**`sync_fleet_update_script()` becomes obsolete.** It exists to publish `/opt/fleet/update.sh` from the image because nothing else refreshed host scripts. Greengrass delivers artifacts per deployment, which solves that problem properly.

**The `last_pull` breadcrumb needs re-pointing.** It writes to `/opt/fleet`, which `greengrass-compose.yaml` does not mount. The equivalent under Greengrass is the Device Shadow — see the observability section of #397, which already proposes reporting disk usage and maintenance outcomes there.

The durable parts are the *findings*, not the code: `ghcr.io` publishes no AAAA record while the blob host does; a v6-only pull only completes where NAT64/DNS64 exists; and the transport-vs-auth error classification (with the regression test for digests containing `401403404`) is worth keeping wherever retry logic ends up living.

Keeping this open until the rework lands so the analysis and tests are not lost.

PR #398, opened 2026-08-03
perf(build): cache CI layers, multi-stage the API image, drop dead deps — 889 MB → 497 MB (#348)

nana-acorn:
Implements #348. Drains `epic/fleet-deployment-hardening` into `main` — the epic currently contains only this work.

## Why now

This is tested and independent of the Greengrass migration: it changes how images are *built*,
not how they are deployed. The remaining PRs on this epic (#362, #364, #367, #368) are all
affected by the migration and will be re-based onto `feat/greengrass-migration` separately.
Merging this first empties the epic so that re-basing is clean.

## The problem

Every OTA shipped the whole **~889 MB** image to every Sentri even for a one-line change, and CI
took a flat ~7 minutes regardless of what changed. One cause for both: no Docker layer cache in
CI, so `pip install` re-executed on every run and produced a fresh layer digest. Watchtower
dedupes on digest, so a new digest means a full re-pull fleet-wide.

## What's in it

**`9a17971` — CI layer cache + multi-stage API image → 889 MB to 497 MB.**
`cache-from`/`cache-to` (type=registry, `:buildcache`) on both image builds, so unchanged
dependencies reuse the layer and keep its digest — pushes, device pulls and on-device storage all
dedupe. **A code-only change now ships ~4 MB.** `Dockerfile.api` splits into builder + runtime:
`build-essential` exists only to compile `RPi.GPIO` and `spidev`, which publish wheels for 32-bit
ARM only while the fleet runs cp311/aarch64, so the wheels are built in the builder stage and
kept out of the runtime image.

**`ff0490e` — remove the dead `simple_rpc` Arduino layer.** That hardware design is gone: motors
run off the Pi's pins via `RPi.GPIO`, the TEC controller over plain pyserial, and fleet compose
maps `/dev/ttyUSB0`, not `ttyACM0`. The library has been *unimportable* for some time —
`simple_rpc` 1.0.5 uses `inspect.getargspec`, removed in Python 3.11 — and nobody noticed, which
is consistent with nothing calling it. Verified identical inside the running `aquila-backend`
container on sn03, so this reflects the deployed state rather than a build artefact.

**`576e788` — publish arm64 only.** The fleet is entirely 64-bit Pi OS. The amd64 variant was
never a decision: before `5db2885` the workflow had no `platforms:` line, so buildx built only
the runner's native platform and the published images could not run on the Pis at all. Adding
arm64 fixed that; amd64 was the implicit default carried forward. Every doc in the repo specifies
arm64 only, and no consumer was found.

**`54331c0` — exclude `venv/` from the build context.** `.dockerignore` excluded `.venv/` but the
repo's own virtualenv is `venv/` — 448 MB of host-architecture packages that a local
`docker compose build` copied straight into an ARM device image. CI was unaffected (`venv/` is
gitignored), so this only ever bit local builds.

Also present: a test commit and its revert (`5944146`, `4f8b2fb`), used to measure a device pull
between two multi-stage images. They cancel out.

## Verification

Sizes measured from a clean `git archive HEAD` export on `linux/arm64` to match CI. The
dedupe claim is the load-bearing one — an unchanged dependency layer must keep its digest across
builds, since that is what makes a code-only change ship ~4 MB instead of the whole image.


nana-acorn:
Superseded by #399 — same commits, on a clearly-named `perf/api-image-size` branch instead of the epic branch, so `epic/fleet-deployment-hardening` can be retired once the remaining PRs are retargeted.

PR #399, opened 2026-08-03
perf(build): cache CI layers, multi-stage the API image, drop dead deps — 889 MB → 497 MB (#348)

nana-acorn:
> **Re-opened against `main`.** This is the same work as #356, which merged into `epic/fleet-deployment-hardening` rather than `main`. The epic is being retired: its remaining PRs move to `epic/fleet-hardening-2` (based on `feat/greengrass-migration`), while this build work is independent of the Greengrass migration — it changes how images are *built*, not how they are deployed — so it goes to `main` on its own. Branch renamed to `perf/api-image-size` to say what it is. Original description follows unchanged.

---

Implements items 1–3 and 5 of #348, plus a dead-code removal found while verifying them.

## The problem

Every OTA shipped the whole **~889 MB** image to every Sentri even for a one-line change, and CI took a flat ~7 min regardless of what changed (measured: 6.8, 7.1, 7.1, 7.2, 7.4, 7.5, 7.5, 7.7).

Both symptoms had one cause: **no Docker layer cache in CI.** Runners are ephemeral, so buildx started cold every run and re-executed `pip install`. pip output is not byte-reproducible (timestamps, `.pyc`, wheel metadata ordering), so that 419 MB layer got a **fresh digest every build**. Watchtower dedupes on digest, so a new digest means a full re-pull fleet-wide.

## Result

| | Layer sum (linux/arm64) |
|---|---|
| Before | 889 MB |
| + multi-stage, − pandas | 583 MB |
| + − simple-rpc | **497 MB** |

**−392 MB (44%).** And with the cache in place, a code-only change ships ~4 MB instead of 889 MB.

## Changes

**1. CI layer cache** — `cache-from`/`cache-to` (`type=registry`, `:buildcache`) on both image builds. Unchanged dependencies now reuse the layer and keep its digest, so pushes, device pulls, and on-device storage all dedupe against what they already have.

`type=registry` over `type=gha`: the GHA cache caps at 10 GB and multi-arch (amd64 + arm64) with a 419 MB pip layer per arch would crowd it, and eviction means silent cache misses.

**2. Multi-stage `Dockerfile.api`** — `build-essential` exists solely to compile `RPi.GPIO` and `spidev`: both publish wheels only for 32-bit ARM (`linux_armv6l`/`armv7l`, cp27–cp310) while the fleet runs cp311/aarch64, so neither matches and pip falls back to the sdist. Wheels are now built in a builder stage and installed with `--no-index`, so a missing wheel fails loudly at build time rather than silently pulling an sdist that cannot build without a compiler.

`curl` is retained in the runtime stage — `entrypoint.sh` uses it to poll `/health` before starting the hardware app.

**3. Dropped `pandas`** — nothing in the backend or hardware path imports it. Its only users are `scripts/tools/` and the Streamlit dashboards in `aquila_web/stream/`, which build from `requirements-ui.txt` (already lists pandas).

**4. Removed the dead `simple_rpc` Arduino layer** — found while verifying the image.

**5. Publish `linux/arm64` only.** The fleet is entirely 64-bit Pi OS. amd64 was never a decision: before `5db2885` the workflow had no `platforms:` line, so buildx built only the runner's native platform (amd64) and the images could not run on the Pis. That commit's fix was *adding* arm64; amd64 was the pre-existing default, carried forward. Every doc in the repo specifies arm64 only.

No consumer found — `compose.yaml` builds locally rather than pulling, fleet compose targets arm64, `docker/docker-compose.yml` is marked unused in its own header, and no workflow pulls or runs the image.

Halves registry storage and build-cache size. It does **not** halve build time: the runner is amd64, so amd64 was the *native* (fast) half and the emulated arm64 half is what remains. QEMU setup is deliberately retained for that reason.

## On the simple_rpc removal

`simple_rpc` drove an Arduino over serial RPC on `/dev/ttyACM0`. That hardware design is gone: motors run straight off the Pi's pins via `RPi.GPIO` (`motor_class.py`), the TEC controller over plain pyserial (`meerstetter.py`), and the fleet compose maps `/dev/ttyUSB0` — only the legacy and local-dev compose files still reference `ttyACM0`.

It has also been **unimportable for some time and nobody noticed**, which is consistent with nothing calling it. `simple_rpc` 1.0.5 uses `inspect.getargspec`, removed in Python 3.11:

```
ImportError: cannot import name 'getargspec' from 'inspect'
```

Confirmed identical inside the running `aquila-backend` container on sn03 — so this is the deployed state, not a build artefact.

Removed `aq_lib/hw_api.py`, `aq_lib/tecControl.py` (already described as a *"legacy example"* in the docs), and `aquila_web/hardware.py`. Nothing in the codebase imported them; only docs referenced them, updated here.

`config_module`'s `find_by_serial_number` / `find_by_vid_pid` are left in place — they are generic Config helpers and tests stub them as part of that interface.

## What left the image

Installed packages: **72 → 46**.

```
Automat  Incremental  MarkupSafe  Pygments  Tubes  Twisted  Werkzeug
attrs  autobahn  cbor2  constantly  flake8  hyperlink  iniconfig  klein
mccabe  msgpack  pandas  pluggy  pycodestyle  pyflakes  pytest
simple_rpc  txaio  ujson  zope.interface
```

**25 of these 26 arrived via `simple-rpc`**; only `pandas` was a direct entry in `requirements-backend.txt`. (It appears here because this diff spans both commits — pandas was not a simple_rpc dependency.)

Worth noting what that list contains: **Twisted, Werkzeug, klein, autobahn** — an entire async web framework stack — plus **flake8, pycodestyle, pyflakes, mccabe, pytest**, declared as hard install dependencies rather than dev extras. Every Sentri in the field has been shipping a linter and a test runner in its production image, pulled in by a library that could not be imported.

## Verification

Built on `linux/arm64` from a clean export to match CI.

| Check | Result |
|---|---|
| Container starts, `/health` responds | `{"status":"ok"}` |
| Docker HEALTHCHECK reaches `healthy` | ✓ — compose and Watchtower depend on this |
| `/version`, `/update/status`, `/results/status`, `/profiles` | all respond |
| `entrypoint.sh` runs, copies 9 bundled profiles | ✓ |
| `aquila_web.main` imports with the 3 modules removed | ✓ |
| All runtime deps import | ✓ |
| `RPi.GPIO` + `spidev` transfer as compiled wheels | ✓ |
| No compiler in runtime image (`which gcc cc make`) | nothing |
| `curl` present for `entrypoint.sh` | ✓ |
| `twisted` / `pytest` absent | ✓ |
| UI image builds | 77.2 MB |
| `compose.yaml` valid | ✓ |
| Workflow YAML parses, cache keys + platforms correct | ✓ |
| Image size | 889 → 497 MB |
| Cache mechanism | 9/13 layers preserved, **0.92 MB delta** |
| Test suite | unchanged vs clean `main` |

### Cache mechanism, demonstrated

Built with `--cache-to`, appended one comment line to `aquila_web/main.py`, rebuilt with `--cache-from`, then compared layer digests between the two OCI outputs:

```
layers identical : 9/13
bytes a device would download for this change: 0.92 MB
```

The 9 preserved layers include every large one — the 74.7 MB pip layer, 54.3 MB base, 28.1 MB, 15.9 MB. Only the four code layers changed. **A one-line edit produces a 0.92 MB delta rather than a full re-pull.**

This ran locally rather than in Actions, but the mechanism is platform-independent: cache → layer digest preserved → registry, device and local store all skip what they already have.

### Tests

**37 failed / 842 passed / 237 skipped — identical with and without these changes**, verified by running the full suite against a stashed clean tree. Those 37 are pre-existing hardware-test pollution failures (they pass in isolation: 87/87) and are unrelated to this work.

### ⚠️ Not exercised by this PR

`docker-build.yml` triggers only on `push: branches: [main]` and `workflow_dispatch` — there is no `pull_request` trigger. **So opening this PR runs no CI**, and the `cache-from`/`cache-to` and `platforms` edits are unverified by real Actions until they reach `main`. The YAML parses and the cache mechanism is proven locally, but buildx has not accepted this exact config in a real run.

It can be tested pre-merge via `workflow_dispatch` against this branch, run twice — the first populates `:buildcache`, the second should be dramatically faster. That does push `:latest` and `:<sha>` from a non-main branch; devices use ring tags so nothing in the fleet moves.

## Not included

- **Item 4 (native arm64 runner)** — deferred, [reasoning here](https://github.com/AcornGenetics/aquilla-main/issues/348#issuecomment-5107304417). Its value shrinks once caching lands, and its economics expire when the repo goes private.
⚠️ **One caveat on item 5.** `docs/testing/testing-pipeline.md` documents a `scripts/test_candidate.sh` that pulls and *runs* the image. That script does not exist in the repo, but if it is implemented and run on an Intel machine or amd64 CI runner it will need an amd64 variant. Apple Silicon is unaffected.

## Known doc debt

`specs/local-db-spec.md` names `hw_api.py` as the caller for `insert_run()` / `close_run()`, and gives instructions to modify it (e.g. *"replace `generate_runid()` in `hw_api.py` with `insert_run()`"*). That file no longer exists after this PR.

Left alone deliberately — it is a design doc, partially implemented via `aquila_web/local_db.py`, and rewriting someone's spec is out of scope here. Flagging so whoever owns that spec can decide whether those integration points need re-pointing or were already superseded.

## Note for the first merge

The first build after this lands is a **cold build** — there is no `:buildcache` tag to restore from yet, so expect the usual ~7 min. The gains start on the second build.

A `:buildcache` tag will appear in the packages list alongside `:prod` and `:latest`. That is expected, and it must be excluded from any future image-retention prune — deleting it silently reverts this fix with nothing obviously broken to point at.

Part of #348 — items 1–3 and 5. Item 4 (native arm64 runner) remains deferred, so this deliberately does not auto-close the issue.


NicoleCornell8 (commented):
This is pointed to main not green grass? this should be done on the green grass application correct?

nana-acorn:
Superseded — these commits are already contained in `epic/fleet-hardening-2` (they landed via `epic/fleet-deployment-hardening`, which epic2 is stacked on). `git rev-list --count origin/epic/fleet-hardening-2..origin/perf/api-image-size` is 0, so there is nothing left for this PR to contribute. The work will reach `main` with the epic2 PR.

Issue #401, opened 2026-08-03
Certificate renewal depends on a throwaway image from the wrong registry, pinned at provisioning


nana-acorn:
Found while implementing #397. Not currently broken — but it is one deleted image away from a silent certificate expiry, and it constrains image cleanup fleet-wide.

## What it does

`aquila-cert-renew.timer` fires daily and renews the device certificate — the credential
proving a device is a genuine Sentri when talking to the cloud. It runs the renewal in a
throwaway container:

```
ExecStart=/usr/bin/docker run --rm \
    -v /opt/aquila/config:/config \
    ghcr.io/acorngenetics/aquilla-main-api:dev \
    python -m aq_lib.renew /config
```

## Current state on sn01 (a migrated Greengrass device)

It works:

```
Aug 03 10:16:26 sn01 systemd[1]: Starting aquila-cert-renew.service...
Aug 03 10:16:28 sn01 docker[9739]: certificate not yet due for renewal — no action
Aug 03 10:16:29 sn01 systemd[1]: Finished aquila-cert-renew.service.

notAfter=Aug 13 21:29:42 2026 GMT   (~10 days out)
```

So this is a latent fault, not an outage. Three things make it fragile.

## 1. It depends on an image nothing holds

`--rm` removes the container the instant it finishes, so for 23h59m a day that image is
unused by Docker's definition. Any cleanup that removes unused images deletes it.

That is not hypothetical: it is the sole reason #397 could not use the one-line
`docker image prune -a` and had to implement a "keep the newest image per repository"
rule instead. The retention rule on every device is shaped around protecting this one
image. Fix this, and that script can be replaced by a built-in.

## 2. It points at the wrong registry

It pulls from **GHCR** on a device that has migrated to **ECR** and now authenticates to
AWS via the token exchange service. The image happens to still be on disk from before the
migration, which is the only reason this works. If it is ever removed, the re-pull needs
GHCR credentials a migrated device may no longer hold.

## 3. Its tag is frozen at provisioning

`${IMAGE_TAG}` was interpolated once, when `deployment2.sh`/`deployment3_greengrass.sh`
wrote the unit — baked as `:dev` on sn01. Under Greengrass nothing maintains those `.env`
values, so the reference is pinned to whatever was true on setup day, permanently.

## Why the failure would be silent

Renewal fails → nothing alerts → the certificate expires → mTLS fails → the device drops
off the fleet, weeks later, with no obvious link back to a missing image. The device also
[...]

PR #406, opened 2026-08-04
fix(clock): install fake-hwclock during Greengrass provisioning (#353)

nana-acorn:
Part of #353 — the host half. The application half (`clock_health.py` reporting an untrustworthy clock) ships in the image and needs nothing from provisioning.

## Why

A Pi has **no battery-backed clock**. On every power loss it resets and stays wrong until NTP resyncs. Two consequences:

**1. Corrupted warehouse data.** Every Event on the sync outbox is stamped with the local clock, so a device that boots wrong writes wrong timestamps into analytics — permanently, and indistinguishable from real data afterwards. This is what #353 was filed for.

**2. Broken authentication.** TLS validates certificates against system time. A device that boots believing it is outside its own certificate's validity window **rejects its own credential** and authenticates to neither AWS IoT nor acorn-ca. It cannot renew its way out either, because renewal authenticates using that same certificate. With 14-day certificates the margin is thin.

The second only became obvious while investigating #401 today, and it makes this the cheapest mitigation available for that whole class of failure.

## Measured, not assumed

Checked sn01 and sn08 directly:

```
fake-hwclock installed: 0        ← the actual gap
NTP enabled:            yes      ← already true
timesyncd active:       active   ← already true
clock synchronized:     yes
```

So of the three commands in the original #367 proposal, **only the package install is a real gap.** The other two are no-ops on existing devices; they are included so a fresh install is deterministic rather than inheriting whatever the OS image defaulted to.

## Why it has to be provisioning

Installing a package must happen on the host, and `fake-hwclock` must run in the boot sequence **before Docker exists** — it restores the clock at boot and saves it at shutdown. A container can do neither. So unlike config-file changes, this genuinely cannot ride the image, and Phase 1 of `deployment3_greengrass.sh` is where every device will run it during migration.

`deployment2.sh` is unchanged; devices that never migrate keep their current behaviour.

## Safety

The enable commands are `|| true` so an unusual image cannot fail provisioning — with `run_test` assertions immediately after, since otherwise a silently unconfigured clock would pass provisioning and nobody would know until a device wrote bad timestamps.

## Tests

3 new static assertions. Suite: 55 pre-existing failures unchanged, +3 passing.

## Verification

After migrating a device: `dpkg -l fake-hwclock`, `systemctl is-enabled fake-hwclock`, and `timedatectl`. The real test is a power-cycle with the network unplugged — the clock should come back approximately right instead of at epoch.

### DIFF
--- a/scripts/deploy/deployment3_greengrass.sh
+++ b/scripts/deploy/deployment3_greengrass.sh
@@ -113,7 +113,38 @@ DEBIAN_FRONTEND=noninteractive apt-get install -y \
     chromium openbox \
     xserver-xorg x11-xserver-utils xinput \
     python3-gi gir1.2-gtk-3.0 gir1.2-webkit2-4.1 libwebkit2gtk-4.1-0 \
-    xterm unclutter
+    xterm unclutter \
+    fake-hwclock
+
+# ── Device clock (#353) ───────────────────────────────────────────────────────
+# A Pi has no battery-backed RTC, so on every power loss the clock resets and stays
+# wrong until NTP resyncs. Two consequences, and the second is the serious one:
+#
+#   * Every Event on the sync outbox is stamped with the local clock, so a device
+#     that boots wrong writes wrong timestamps into the analytics warehouse —
+#     permanently, and indistinguishable from real data afterwards.
+#
+#   * TLS checks a certificate's notBefore/notAfter against system time. A device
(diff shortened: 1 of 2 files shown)


NicoleCornell8 (changes requested):
this should all point to green grass feature no?

Issue #397, opened 2026-08-03
Greengrass: cap on-device image retention from the component lifecycle


nana-acorn:
> **Updated 2026-08-05 to reflect what shipped.** Image retention is implemented and merged
> (`scripts/greengrass/prune-images.sh`, in `epic/fleet-hardening-2`). Two things changed from
> the original proposal, both to avoid needing root:
>
> * **There is no separate maintenance component.** The prune ships as an artifact of the
>   existing app component and runs from its `Run` lifecycle as `ggc_user`. The root question
>   below is therefore *not* live for this issue, and needs no sign-off.
> * **Host-config convergence (the old Responsibility 2) is removed from this issue.** It was
>   the only part that required root. It remains unsolved and is still tracked by #353 and
>   #354; nothing here delivers it.
>
> The original framing is kept below where it is still accurate, with corrections marked.

Supersedes the approach in #362 (now **closed** in favour of this). Implements the intent of
#355 for Greengrass.

## Problem: images accumulate forever

Nothing on a device has ever removed a container image. Measured live in the Watchtower world:

| Device | Ring | Images | Untagged | Reclaimable | Disk |
|---|---|---|---|---|---|
| **sn03** | sandbox | 30 | **18** | ~9.8 GB (82%) | 40% of 58 GB |
| **sn07** | pilot | 8 | 5 | 2.32 GB (65%) | 22% of 58 GB |

The oldest image on sn03 is `timberio/vector` from 2023-10-30 — part of a telemetry stack
replaced by Grafana Alloy and never cleaned up.

**Confirmed on Greengrass too.** sn01, a live Greengrass device, was measured after a second
deployment still holding the previous deployment's 1.17 GB ECR image alongside the current one,
plus 4.85 GB of older dangling layers — ~7.6 GB reclaimable, and roughly another 1.25 GB set per
deployment indefinitely. Greengrass's `DockerApplicationManager` pre-stages a fresh digest per
deployment and never collects the old one.

#362 addressed this for the current Watchtower/GHCR setup. That approach does not survive the
Greengrass migration, and fails **silently** rather than loudly:

1. **Ring detection dies.** The script derives its ring from `IMAGE_TAG` in `/opt/fleet/.env`.
   Greengrass deploys component *versions*, not ring tags, so the variable is meaningless. The
   script hits `exit 0` — "refusing to remove anything" — while `systemctl is-active` still
   reports healthy. Nothing is ever deleted and nothing looks wrong.
[...]

Issue #363, opened 2026-07-28
Greengrass migration: extend cert renewal to rotate the IoT cert + reconnect Greengrass


NicoleCornell8:
## What to build

Extend the device-certificate **renewal** path so that when the short-lived `acorn-ca`
leaf renews, it **also rotates the IoT/Greengrass connection cert and forces Greengrass
to reconnect** — not just the file used for analytics mTLS. Analytics mTLS is
request-scoped (short TTL is fine), but IoT/Greengrass holds a **long-lived MQTT
connection** that must pick up the renewed cert or it eventually drops off the control plane.

**Target branch:** sub-branch off `feat/greengrass-migration`.

## Acceptance criteria

- [ ] On renewal, the IoT connection cert is rotated (not only the analytics cert)
- [ ] Greengrass reconnects cleanly with the renewed cert (no dropped control-plane connection)
- [ ] Renewal stays a no-op until the cert passes its renewal threshold (unchanged behavior)
- [ ] Verified on the pilot device (rotation does not drop the Greengrass connection)

## Blocked by

- AcornGenetics/aquilla-main#361

PR #408, opened 2026-08-05
merge: fleet-hardening epic into the Greengrass migration branch (#348, #397)

nana-acorn:
Brings the fleet-hardening epic into the Greengrass migration branch. Two independent
strands, both already reviewed on their own PRs:

**Build/image size (#348)** — CI layer caching, a multi-stage API image, arm64-only
publish, and removal of the dead `simple_rpc` Arduino layer and pandas. 889 MB → 497 MB,
which is what a device actually pulls on every update.

**Device image retention (#397)** — `scripts/greengrass/prune-images.sh` caps retained
images from the component lifecycle, and collects dangling images too rather than only
the ones it listed. Without it a device accumulates every image it has ever pulled.

Merges cleanly (verified with `git merge-tree`). No conflicts against the 4 commits
`feat/greengrass-migration` has picked up since the merge-base (#396, #402, #403, #407).

Note: #369 (OTA pull retry + IPv6 fallback) is still open against
`epic/fleet-hardening-2` and is *not* included here — it will flow in via a later merge.


docker/Dockerfile.api
```diff
@@ -1,16 +1,39 @@
...
-    build-essential \
     procps \
     sudo \
     curl \
     && rm -rf /var/lib/apt/lists/*
 
+# --no-index: install only from the wheels built above, never from PyPI. If a
+# wheel is missing this fails loudly at build time rather than silently pulling
+# a compiler-free sdist that cannot build.
+COPY --from=builder /wheels /wheels
```
NicoleCornell8: **Priority:** `rm -rf /wheels` on line 36 doesn't shrink the image. This `COPY --from=builder /wheels /wheels` creates its own layer holding all the wheels (numpy/matplotlib/cryptography/awscrt ≈ 150–250 MB); deleting them in the later `RUN` only whites them out — the bytes stay in this COPY layer and ship to every device, so the runtime image is ~a full package-tree larger than intended (undercuts the 889→497 MB goal).

Fix: install from a BuildKit bind mount so nothing lands in a persisted layer:
```dockerfile
RUN --mount=type=bind,from=builder,source=/wheels,target=/wheels \
    pip install --no-cache-dir --no-index --find-links=/wheels -r requirements-backend.txt
```
(drop the standalone `COPY --from=builder` + `rm -rf /wheels`).

scripts/greengrass/recipe.py
```diff
@@ -75,16 +104,18 @@ def build_recipe(component_name, version, compose_artifact_uri, image_refs=None)
                         # this process env (Greengrass sets them) and compose passes
                         # them through; the thing name is not, so export it here.
                         'export AWS_IOT_THING_NAME="{iot:thingName}"; '
-                        "docker compose -f %s up -d; "
+                        + ("docker compose -f %s up -d; " % compose_path)
+                        # reclaim images the deployment just superseded (#397)
+                        + prune_step
```
NicoleCornell8: The prune is chained with `;` after `up -d`, so it runs even when `up -d` fails. On a redeploy, Shutdown has already `docker compose down`'d the previous containers, so if the new `up -d` fails for a non-pull reason (bad compose, port clash) nothing references the previous-good image — and if the freshly-staged new image is newest-per-repo, the old good image is neither protected nor newest and gets pruned, so rollback has to re-pull it.

Cheap hardening: gate the prune on success — `docker compose ... up -d && bash "{artifacts:path}/prune-images.sh" || true` — so cleanup only runs once the new stack is actually up.

scripts/greengrass/prune-images.sh
```diff
@@ -0,0 +1,171 @@
...
+    while IFS= read -r id; do
+        [[ -z "${id}" ]] && continue
+        if [[ "${DRY_RUN}" == "1" ]]; then
+            log "would remove ${id}"
+            removed=$((removed + 1))
+            continue
+        fi
+        # Failure is expected and fine: docker refuses to remove an image another
+        # image layers on, and that is a floor beneath this script's own logic.
+        if docker rmi "${id}" >/dev/null 2>&1; then
```
NicoleCornell8: `docker rmi <id>` fails for any image carrying more than one repository tag (`must be forced - image is referenced in multiple repositories`). CI pushes three tags (`-api:latest`, `-api:<sha>`, and optionally `:<ring_tag>`), so a migrated Watchtower-era device that pulled an image under >1 of them will log "could not remove" and retain exactly the stale GHCR image #397 is meant to shed. `test_a_migrating_device_sheds_its_old_ghcr_images` only covers single-tag (`:dev`) images, so this path is untested.

Low priority (it fails safe — retains rather than mis-deletes), but consider `docker rmi -f` here since the newest/protected guards have already excluded live images.

.github/workflows/docker-build.yml
```diff
@@ -57,13 +60,21 @@ jobs:
...
             ghcr.io/${{ env.REPO_LOWER }}-api:${{ github.sha }}
             ${{ inputs.ring_tag && format('ghcr.io/{0}-api:{1}', env.REPO_LOWER, inputs.ring_tag) || '' }}
+          # Runners are ephemeral, so without an external cache buildx starts cold every
+          # run and re-executes `pip install`. pip output is not byte-reproducible
+          # (timestamps, .pyc, wheel metadata ordering), so that layer got a fresh digest
+          # every build — meaning devices re-pulled the whole ~889 MB image for a one-line
+          # change. Reusing the layer preserves its digest, so pushes, device pulls and
+          # on-device storage all dedupe against what they already have. See #348.
+          cache-from: type=registry,ref=ghcr.io/${{ env.REPO_LOWER }}-api:buildcache
+          cache-to: type=registry,ref=ghcr.io/${{ env.REPO_LOWER }}-api:buildcache,mode=max
```
NicoleCornell8, line 70: GHCR can reject buildx's default registry-cache manifest media types, so `cache-to: type=registry` may fail to export the cache — which would silently negate the layer-digest-preservation this PR relies on (devices go back to re-pulling full images). Please confirm the first CI run actually exported it (look for a cache-export step in the build log, not just cache-import). If it didn't, the robust form is:
```yaml
cache-to: type=registry,ref=...:buildcache,mode=max,image-manifest=true,oci-mediatypes=true
```
(same for the UI cache on line 87.)

NicoleCornell8 (commented):
Review of the fleet-hardening merge. One priority fix (the wheels layer), the rest are lower-severity — flagged inline. Deletions (`hw_api`/`tecControl`/`hardware`/`simple_rpc`, and `pandas` from the backend) are clean — no remaining references. The prune retention logic is solid and well-tested.

nana-acorn:
## Update — review feedback merged into the epic

Both follow-ups from @NicoleCornell8's review are now in this branch: #409 (`e526d18`) and #411 (`a6db05a`).

### #409 — prune only once the deployment is healthy

The prune ran immediately after `docker compose up -d`, before the `/health` gate. At that point the previous image is referenced by nothing and has lost "newest per repository" to the image just deployed — so it was deleted, and that is precisely the image Greengrass rolls back to when the new version then fails `/health`. Cleanup was destroying the rollback target. Now it runs after the gate.

### #411 — wheels layer bloat and the multi-tag `rmi`

**Wheels layer.** `COPY --from=builder /wheels` wrote its own layer; the later `rm -rf` only whited the bytes out, so they shipped to every device. Replaced with a BuildKit bind mount, which is never committed. Note this means **BuildKit is now required** — `DOCKER_BUILDKIT=0 docker build` fails with `unknown flag: mount`. CI and modern `compose build` are unaffected; documented in `Dockerfile.api`.

**Multi-tag removal.** `docker rmi` refuses an image carrying two tags, which is the routine shape on a migrated device since CI pushes `:latest` and `:<sha>`. Without a retry the script kept exactly the stale images it exists to shed. Now retried with `-f` for that conflict only.

**And the follow-up finding, which was the important one.** The first version of that retry keyed on `"must be forced"`, which is broader than the multi-tag case. Verified against docker 28.4.0:

```
$ docker run --name t hello-world && docker rmi <id>
conflict: unable to delete 7f4da0fc94bc (must be forced)
  - image is being used by stopped container 0382f5e9e424
$ docker rmi -f <id>     # succeeds — image gone, container orphaned
```

So the guard fired on the exact case it was written to avoid. The protected set does not cover it either: that is a start-of-run snapshot, so a container created afterwards (`aquila-cert-renew` starts one daily) is not in it. Now additionally requires the refusal not to be a container reference.

The test that claimed to cover this omitted docker's `(must be forced)` prefix, so it passed while asserting a property that did not hold. Corrected — it fails against the old guard — plus a standalone regression test.

## Branch state

```
a6db05a  merge #411 — review fixes
e526d18  merge #409 — prune after the health gate
7db5a1e  merge #400 — image retention from the component lifecycle
e769143  merge — #348 build work (889 MB → 497 MB)
```

**This branch is 6 commits behind `feat/greengrass-migration`** and should be merged up before review, so the diff shows only this epic's changes.

## Still open, not in this branch

- **#404** (cert renewal via `docker exec`) — held. It makes renewal more reliable, which given #363 currently means more likely to rotate a certificate AWS has not registered. Wants the rotation/JITP question settled first.
- **#369** (pull retry / IPv6) — to be reworked for Greengrass; most of it does not port, since `DockerApplicationManager` owns the transfer.
- **#406** (`fake-hwclock`) — separate PR, pending a reboot test: systemd's own `/var/lib/systemd/timesync/clock` may already cover the same ground.

PR #411, opened 2026-08-05
fix: address #408 review — wheels layer bloat and multi-tag rmi (#348, #397)

nana-acorn:
Addresses @NicoleCornell8's review of #408. Two of the four comments needed code; the other
two are answered below.

## 1. Wheels layer (priority) — fixed

`COPY --from=builder /wheels /wheels` writes its own layer holding every wheel; the later
`rm -rf /wheels` only whites them out, so the bytes ship to every device. Replaced with a
BuildKit bind mount, which is never committed:

```dockerfile
RUN --mount=type=bind,from=builder,source=/wheels,target=/wheels \
    pip install --no-cache-dir --no-index --find-links=/wheels -r requirements-backend.txt
```

Added a `# syntax=docker/dockerfile:1.4` directive so the mount syntax is understood
regardless of the builder's default. Verified with `docker buildx build --check` — no warnings.

Note the 889 → 497 MB figure was measured, so it already included this bloat; the fix makes
the image smaller still rather than invalidating the number.

## 2. `docker rmi` multi-tag — fixed, and confirmed worse than reported

Verified against docker directly:

```
$ docker tag hello-world testrepo/x:latest && docker tag hello-world testrepo/x:v2
$ docker rmi <id>
Error response from daemon: conflict: unable to delete 7f4da0fc94bc (must be forced)
  - image is referenced in multiple repositories
```

The message says "multiple repositories" but it fires on **two tags in a single repository**.
CI pushes `:latest` and `:<sha>` on every build, so this is the common shape on a migrated
device, not an edge case.

Retried with `-f` **only** on that message, not blanket. `docker rmi -f` on an image a stopped
container references removes it and orphans the container, which would dissolve the floor the
script deliberately leans on. The protected set already excludes container-referenced images,
so `-f` is a second line here rather than the first; any other refusal still stops the removal
and is logged. A forced removal says so, since it means the device is carrying multi-tagged
images.

Four tests: forced on the multi-tag conflict, never forced on a container reference, reported
when `-f` also fails, never forced under `DRY_RUN`. The first fails against the unfixed script
with the exact reported symptom (`could not remove old` / `0 removed`).

## 3. Prune chained with `;` after `up -d` — already fixed in #409

#409 (open against this same base) moved the prune after the health poll and gated it on a
`/health` probe. That subsumes the suggested `up -d && prune`: if `up -d` fails for any reason
— bad compose, port clash, pull failure — nothing answers `/health`, so the prune does not run.
Same bug found from two directions.

## 4. Registry cache media types — does not reproduce

Checked the last `docker-build.yml` run:

```
#24 exporting cache to registry
#24 sending cache export 4.4s done
```

Both API and UI exported cleanly; GHCR accepted the default media types. The only failure in
that log is an unrelated Actions cache for `binfmt`. No change made.

## Device verification

The prune ran for real on sn03 (sandbox, Watchtower-era) during this review:

| | Before | After |
|---|---|---|
| Images | 30 | 6 |
| Image size | 11.83 GB | 1.035 GB |
| Reclaimable | 10.7 GB (90%) | 255.4 MB (24%) |
| Disk used | 24 G (43%) | 14 G (25%) |

10 GB genuinely returned to the filesystem, all four containers still healthy, and the three
running images correctly protected. Note the removal set was 22 dangling + 2 single-tag images,
so it did **not** exercise the multi-tag path fixed here — that remains device-unverified.

Full suite: 64 failed / 931 passed / 237 skipped, against an `epic/fleet-hardening-2` baseline
of 64 / 927 / 237 — same pre-existing macOS hardware and matplotlib failures.

### DIFF
--- a/docker/Dockerfile.api
+++ b/docker/Dockerfile.api
@@ -1,6 +1,13 @@
 # syntax=docker/dockerfile:1.4
 # Pins the BuildKit frontend so `RUN --mount=type=bind,from=...` below is understood
 # regardless of the builder's default Dockerfile version.
+#
+# BUILDKIT IS REQUIRED to build this image. Both the syntax directive above and the
+# bind mount below are BuildKit features, so a legacy build --
+#     DOCKER_BUILDKIT=0 docker build ...
+# or a Docker old enough to default to the classic builder -- fails with
+# `unknown flag: mount`. CI (docker/build-push-action) and modern `docker compose
+# build` use BuildKit by default, so the normal paths are unaffected.
 
 # ── Build stage ───────────────────────────────────────────────────────────────
 # build-essential exists solely to compile RPi.GPIO and spidev from source: both
--- a/docker/Dockerfile.api
+++ b/docker/Dockerfile.api
@@ -1,3 +1,7 @@
+# syntax=docker/dockerfile:1.4
+# Pins the BuildKit frontend so `RUN --mount=type=bind,from=...` below is understood
+# regardless of the builder's default Dockerfile version.
+
 # ── Build stage ───────────────────────────────────────────────────────────────
 # build-essential exists solely to compile RPi.GPIO and spidev from source: both
 # publish wheels only for 32-bit ARM (linux_armv6l/armv7l, cp27-cp310), and the
@@ -30,10 +34,15 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # --no-index: install only from the wheels built above, never from PyPI. If a
 # wheel is missing this fails loudly at build time rather than silently pulling
 # a compiler-free sdist that cannot build.
-COPY --from=builder /wheels /wheels
+#
+# Mounted, not COPY'd. A `COPY --from=builder /wheels /wheels` writes its own layer
+# holding every wheel (numpy/matplotlib/cryptography/awscrt ≈ 150-250 MB), and
+# deleting them in a later RUN only whites them out — the bytes stay in the earlier
+# layer and ship to every device. A bind mount exists for the duration of this RUN
+# and is never committed, so the wheels leave no trace in the image at all.
 COPY requirements-backend.txt ./requirements-backend.txt
-RUN pip install --no-cache-dir --no-index --find-links=/wheels -r requirements-backend.txt \
-    && rm -rf /wheels
+RUN --mount=type=bind,from=builder,source=/wheels,target=/wheels \
+    pip install --no-cache-dir --no-index --find-links=/wheels -r requirements-backend.txt
 
 COPY . .
 COPY profiles/bundled/ /opt/aquila/profiles_bundled/
--- a/scripts/greengrass/prune-images.sh
+++ b/scripts/greengrass/prune-images.sh
@@ -167,19 +167,38 @@ main() {
         # logs "could not remove" and keeps exactly the stale GHCR images it exists
         # to shed.
         #
-        # Forced only for that message, never blanket. `docker rmi -f` on an image a
-        # stopped container references removes it and orphans the container; the
-        # protected set already excludes those, so -f here is a second line rather
-        # than the first, and any other refusal still stops the removal.
+        # Forced only for that message, never blanket -- and "must be forced" alone is
+        # NOT specific enough. Verified against docker 28.4.0, a stopped-container
+        # reference produces:
+        #     conflict: unable to delete <id> (must be forced) -
+        #     image is being used by stopped container <cid>
+        # which also contains "must be forced", and `-f` on it succeeds: the image goes
+        # and the container is left pointing at nothing. That is the precise case this
+        # retry must avoid, so the guard additionally requires the refusal NOT to be a
+        # container reference.
+        #
+        # Excluding "being used by" rather than matching "referenced in multiple
+        # repositories" positively: unfamiliar wording from a future docker then means
+        # no force, which is the safe direction to be wrong in. A positive match would
+        # silently stop reclaiming multi-tag images if the phrasing ever changed.
+        #
+        # The protected set already excludes container-referenced images, but it is a
+        # start-of-run snapshot -- a container appearing after it (aquila-cert-renew
+        # runs one daily) would slip through. This guard is what actually holds.
         rmi_err="$(docker rmi "${id}" 2>&1 >/dev/null)" && rc=0 || rc=$?
-        if (( rc != 0 )) && [[ "${rmi_err}" == *"must be forced"* ]]; then
+        if (( rc != 0 )) \
+            && [[ "${rmi_err}" == *"must be forced"* ]] \
+            && [[ "${rmi_err}" != *"being used by"* ]]; then
             rmi_err="$(docker rmi -f "${id}" 2>&1 >/dev/null)" && rc=0 || rc=$?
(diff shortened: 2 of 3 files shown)


scripts/greengrass/prune-images.sh
```diff
@@ -156,7 +156,27 @@ main() {
...
+        # routinely holds images under two tags -- and without the retry the script
+        # logs "could not remove" and keeps exactly the stale GHCR images it exists
+        # to shed.
+        #
+        # Forced only for that message, never blanket. `docker rmi -f` on an image a
+        # stopped container references removes it and orphans the container; the
+        # protected set already excludes those, so -f here is a second line rather
+        # than the first, and any other refusal still stops the removal.
+        rmi_err="$(docker rmi "${id}" 2>&1 >/dev/null)" && rc=0 || rc=$?
+        if (( rc != 0 )) && [[ "${rmi_err}" == *"must be forced"* ]]; then
```
NicoleCornell8: The `*"must be forced"*` match is broader than the multi-tag case it's meant for. Real docker emits, for a **stopped-container** reference:
```
conflict: unable to delete <id> (must be forced) - image is being used by stopped container <cid>
```
That string contains `must be forced`, so this guard is true and line 176 runs `docker rmi -f`, removing the image and orphaning the container — the exact case the comment says `-f` must avoid. The substring match doesn't exclude it; the only thing that does is the start-of-run protected-set snapshot (line 15). If a container appears between that snapshot and the rmi (e.g. `aquila-cert-renew` mid-lifecycle), the "second line of defense" the comment advertises isn't there.

Blast radius is small in practice (newest-per-repo + protected snapshot shield the images that matter), but the guard should key on the specific message:
```bash
if (( rc != 0 )) && [[ "${rmi_err}" == *"referenced in multiple repositories"* ]]; then
```
which matches the multi-tag/multi-repo conflict and not the stopped-container one. (`cannot be forced - dependent child images` is already excluded since it says "cannot", so that floor is fine.)

unit_tests/test_prune_images.py
```diff
@@ -265,3 +265,107 @@ def test_dangling_images_are_enumerated() -> None:
...
+    result = _run(f"""
+        source {SCRIPT}
+        collect_images() {{
+            printf 'new\t2026-08-03T09:35:43Z\tghcr.io/x/api:dev\\n'
+            printf 'old\t2026-01-01T00:00:00Z\tghcr.io/x/api:old\\n'
+        }}
+        protected_ids() {{ printf ''; }}
+        docker() {{
+            if [[ "$1" == "rmi" && "$2" == "-f" ]]; then echo "FORCED" ; return 0; fi
+            echo "conflict: unable to delete old - image is being used by stopped container abc" >&2
```
NicoleCornell8: This mock omits docker's real `(must be forced)` prefix, so the test passes without actually exercising the dangerous path. Real docker prepends it:
```
conflict: unable to delete old (must be forced) - image is being used by stopped container abc
```
With that string, the guard on line 175 keys on `must be forced` and **would** force-remove the image — so the property this test claims to protect ("must not force a container-referenced image") doesn't actually hold. If you change the mock to include `(must be forced)`, this test fails against the current code, which demonstrates the guard finding above. Pairs with the fix suggested there (match `referenced in multiple repositories` instead).

docker/Dockerfile.api
```diff
@@ -1,3 +1,7 @@
+# syntax=docker/dockerfile:1.4
```
NicoleCornell8, line 1: Minor / heads-up: `RUN --mount=type=bind` makes BuildKit mandatory now. CI (`build-push-action`) and modern `docker compose build` default to BuildKit so they're fine, but a legacy path — `DOCKER_BUILDKIT=0 docker build`, or older Docker — that previously built this image will now fail with `unknown flag: mount`. Worth a one-line note here or in the dev docs; no code change needed.

scripts/greengrass/prune-images.sh
```diff
@@ -156,7 +156,27 @@ main() {
...
+            rmi_err="$(docker rmi -f "${id}" 2>&1 >/dev/null)" && rc=0 || rc=$?
+            (( rc == 0 )) && log "removed ${id} (forced: image had several tags)"
+        fi
+        if (( rc == 0 )); then
             removed=$((removed + 1))
         else
             log "could not remove ${id} (still referenced)"
```
NicoleCornell8: Cosmetic: this now also fires when a `must be forced` image's `-f` retry also failed, or for `cannot be forced - dependent child images` — neither of which is "still referenced", so the hardcoded reason can mislead someone debugging why space isn't being reclaimed. Consider logging the captured message instead, e.g. `log "could not remove ${id}: ${rmi_err}"`, so the real docker cause is visible.

NicoleCornell8 (commented):
Thanks for the thorough fixes. The wheels bind-mount is the right fix (nothing lands in a committed layer), and I traced the `-f` retry under `set -euo pipefail` — the `&& rc=0 || rc=$?` and `(( ))` paths are all set-e-safe. One real issue in how the force-retry guard is scoped, plus its test not covering the dangerous case; two lower-severity notes. Details inline.

nana-acorn:
All four addressed in `d8afc5a`. Thanks — the first two were right, and the second one is the more valuable catch.

## 1. Force-retry guard — confirmed and fixed

Reproduced against docker 28.4.0:

```
$ docker run --name t hello-world && docker rmi <id>
conflict: unable to delete 7f4da0fc94bc (must be forced)
  - image is being used by stopped container 0382f5e9e424

$ docker rmi -f <id>     # succeeds
$ docker ps -a           # container still there, image gone
```

So the guard fired on exactly the case its own comment said `-f` must avoid. And your point about the snapshot is the part I had wrong in the comment: `protected_ids` is captured once at start of run, so a container created afterwards — `aquila-cert-renew` starts one daily — is not in it. The guard was the only thing standing there, and it was not standing.

**One deviation from your suggestion.** I excluded `being used by` rather than positively matching `referenced in multiple repositories`:

```bash
if (( rc != 0 )) \
    && [[ "${rmi_err}" == *"must be forced"* ]] \
    && [[ "${rmi_err}" != *"being used by"* ]]; then
```

Both close the hole. The reason for the negative form: a positive match fails *unsafe* if docker ever rewords that message — the retry silently stops firing and the script quietly goes back to keeping the stale images #397 exists to shed, with nothing to notice. Excluding the container case fails *safe*: unfamiliar wording means no force. Happy to switch if you prefer the positive match.

## 2. The test — you were exactly right

Adding `(must be forced)` to the mock made it fail against the old guard, with:

```
prune: removed old (forced: image had several tags)
```

A test that asserts a safety property and cannot fail when that property is violated is worse than no test, since it advertises a guarantee that is not there. Corrected, and added `test_stopped_container_refusal_is_never_forced` as a standalone regression so the mock and the guard can't drift back together.

## 3. BuildKit — noted in the Dockerfile

Added a comment stating it is now mandatory, that `DOCKER_BUILDKIT=0` fails with `unknown flag: mount`, and that CI and modern `compose build` are unaffected. No code change, as you said.

## 4. Log message — fixed

Now logs docker's own message. Added a test asserting a `cannot be forced - dependent child images` refusal surfaces that text rather than "still referenced".

Suite: 55 pre-existing failures unchanged, +2 passing.

Issue #431, opened 2026-08-07
Splash-to-app handoff is glitchy: cross-origin redirect forces a second Chromium renderer


nana-acorn:
## Summary

The handoff from the boot splash to the app is visibly glitchy — a blank gap between the splash disappearing and the app painting, on every boot.

Cause: `splash.html` is loaded from `file://` and redirects to `http://localhost:8090`. That crosses an origin boundary, so Chromium tears down the renderer it just warmed up and builds a second one from scratch for the app.

```js
setTimeout(() => { window.location.href = 'http://localhost:8090'; }, 450);
```

This issue covers **splash → app** only. The white frame before the splash paints is a separate problem — see #428.

## Evidence — measured A/B on sn11

Same page, same flags, only the URL differs. App already running in both runs, so the health poll succeeds immediately and only the handoff is under test.

| Run | Splash served from | Splash → app |
|---|---|---|
| **A** | `http://localhost:8090/static/splash.html` (same origin) | **Seamless** |
| **B** | `file:///opt/aquila/splash.html` (what we ship) | Visible gap |
| Real boot | `file://` | Worst of the three |

Reproduced by copying the existing file into the backend's static directory — no code change:

```bash
docker cp /opt/aquila/splash.html aquila-backend:/opt/aquila/aquila_web/static/splash.html
curl -s -o /dev/null -w '%{http_code}\n' http://localhost:8090/static/splash.html   # 200
```

Note **the port is part of the origin.** Serving the splash from the existing nginx container on 8080 would not help; it has to be 8090, the same port the app is on.

## Second problem the same change fixes

A `file://` page fetching `http://localhost:8090/health` is cross-origin, so the poll only works because Phase 9b launches Chromium with **`--disable-web-security --allow-file-access-from-files`**.

Confirmed by accident: running the `file://` splash *without* those flags hangs forever on the splash, because the health poll is blocked.

Served from `localhost:8090`, the poll is same-origin and simply allowed — so both flags can be dropped. Running a browser with web security disabled on a device that holds device certificates and talks to AWS IoT is worth removing on its own merits, independent of the glitch.

## Proposed fix

Put nginx in front on 8090 and proxy through to the backend:

- **Backend not up yet** — nginx cannot reach upstream, serves `splash.html`.
- **Backend up** — nginx proxies through, the app is served.

[...]

Issue #428, opened 2026-08-06
Kiosk startup: white frame before the Acorn splash paints


nana-acorn:
## Summary

At the start of the kiosk session, before the Acorn splash paints, the screen shows a brief white frame — originally with an Openbox title bar reading **"Untitled — Chromium"**. Roughly 1/12 second, every boot.

Observed sequence: **pitch black → white flash → black → splash → app.**

Customer-facing: it is the first non-black thing on screen, where Acorn branding should be.

Scope: **fleet-wide, current provisioning path.** sn09–sn12 have it. sn03 also has it, visible only after #426 was fixed — the legacy WebKit kiosk was painting over it.

This issue covers **window → splash** only. The splash → app transition is a separate problem with a separate cause and fix; see the companion issue.

## Verified partial fix — the title bar

An Openbox rule removes the decoration:

```xml
<applications>
  <application class="*">
    <decor>no</decor>
  </application>
</applications>
```

**Tested on sn03: title bar gone, nothing else affected.** Stateless config, no runtime moving parts.

Note we currently deploy **no Openbox window rules at all** — no `rc.xml` anywhere in the repo, Phase 9b writes only `autostart` — so Openbox runs stock defaults that decorate every window. The legacy WebKit kiosk guarded against this explicitly (`server_web/kiosk.py:16`, `set_decorated(False)`); the Chromium path relies on `--kiosk` alone and has no equivalent.

Cost: Openbox has no drop-in config directory, so this means copying the distro `rc.xml` and owning a fork of it. Given #424 and #426 were both caused by exactly that pattern, the copy should carry a comment explaining why it exists.

## The white frame — three ruled-out theories

Recorded so nobody retries them:

1. **Chromium's blank page.** `--default-background-color` (with correct 8-digit RGBA) and `--cast-app-background-color` both had no effect.
2. **The X root window.** `xserver-command=X -nocursor -background none` on both LightDM configs changed nothing.
3. **Something Chromium's window covers.** Hiding the window with `xdotool windowunmap` *exposed* white rather than removing it — so it is behind Chromium — but blacking the root with `xsetroot -solid black` did not fix it either, and the xdotool approach introduced a blue border and a small black square while the window was re-mapped at default geometry. Reverted.

Latest description after all reverts: white, drawn progressively — **top half of the screen first, then the whole screen, then the logo.** That reads as Chromium's first paint being visible in chunks, not as a separate surface underneath.

## Proposed direction — Plymouth

[...]

Issue #426, opened 2026-08-06
Old devices launch two browsers at startup: legacy system-wide Openbox autostart still active


nana-acorn:
## Summary

Older devices launch **two browsers** at startup, not one. The second is the legacy WebKit kiosk, which paints its default error page over the Acorn splash before the backend is listening:

> `Error receiving data: Connection reset by peer`

Plain unstyled text, top-left of a white page — WebKitGTK's default error page, not Chromium's (Chromium renders a styled, centred page).

Customer-facing: it is on screen during startup, before the app appears.

## Root cause

Openbox reads **two** autostart files and runs **both**:

1. `/etc/xdg/openbox/autostart` — system-wide, written by `deployment1.sh`
2. `~/.config/openbox/autostart` — per-user, written by Phase 9b

Phase 9b writes only the second. It never touched the first, so the per-user file **adds** the Chromium kiosk rather than replacing the legacy one.

Confirmed on sn03, both running simultaneously:

```
$ pgrep -af kiosk.py
2129 python3 /home/pi/kiosk.py

$ pgrep -af chromium
2270 /usr/lib/chromium/chromium ... --kiosk file:///opt/aquila/splash.html
```

`server_web/kiosk.py` loads `http://localhost:8090` directly with no retry and no error handling, so any load before the backend is up renders WebKit's error page.

## Three regressions from one file

The legacy `/etc/xdg/openbox/autostart` on sn03:

```bash
xrandr --output HDMI-2 --rotate left
unclutter -idle 0.5 &
sleep 3
python3 /home/pi/kiosk.py &
```

1. **`kiosk.py`** — the error page above.
2. **`xrandr --rotate left`** — fights Phase 9b's `--rotate right`; last writer wins.
3. **`unclutter`** — which Phase 9b explicitly refuses to run. Its own comment: *"it only hides on an idle timer and re-shows the cursor on pointer/touch events, causing a startup flash and a lingering cursor over dropdowns on every tap."* So the cursor behaviour that `specs/hardware/mouse-cursor-removal.md` set out to fix is still happening on these devices.

## Verified fix

On sn03, overwriting `/etc/xdg/openbox/autostart` with an empty file cleared the error entirely. Only the Chromium kiosk remained.

**Overwrite, not delete:** `openbox` ships this path as a dpkg conffile. A deleted conffile can be restored on package upgrade; a modified one is kept.

## Affected devices

[...]

Issue #424, opened 2026-08-06
Console text flashes on screen before the kiosk: legacy console autologin still active alongside LightDM


nana-acorn:
## Summary

Devices carry **two competing autologin mechanisms**. On boot, the legacy one prints a block of console text across the display before the kiosk appears:

```
login: pi (automatic login)
Linux sn03 6.12.75+rpt-rpi-08 #1 SMP PREEMPT Debian ... aarch64
The programs included with the Debian GNU/Linux system are free software; ...
Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent permitted ...
X.Org X Server 1.21.1.16
```

Like #418, this is customer-facing: it is on screen before any Acorn branding. Unlike #418 it is a **userspace** problem, not the bootloader — different stage, different cause, and the two should not be conflated.

## Root cause

`deployment1.sh` set up **console autologin**: an `agetty --autologin pi` override on tty1, plus `startx` in `~/.bash_profile` (`server_web/.bash_profile`).

`deployment2.sh` / `deployment3_greengrass.sh` Phase 4 replaced that with **LightDM autologin** (`do_boot_behaviour B4`, `autologin-session=openbox`).

**Phase 4 sets up the new path but never tears down the old one**, and the two are not mutually exclusive. Any device provisioned before Phase 4 existed still has both.

Confirmed on sn03:

```
$ cat /etc/systemd/system/getty@tty1.service.d/*.conf
[Service]
ExecStart=
ExecStart=-/sbin/agetty --autologin pi --noclear %I $TERM

$ systemctl is-active lightdm
active
```

With both present, agetty logs in on tty1 and prints the banner, MOTD and X.Org version block to the display before LightDM takes the screen. `~/.bash_profile` also races LightDM to start a **second X server** on VT1.

Note the cmdline is already correct (`console=tty3 quiet ... vt.global_cursor_default=0`), so this is not kernel output and `console=tty3` cannot suppress it — a login shell is being started on the console deliberately.

## Verified fix

On sn03, removing both cleared the text entirely with no effect on the kiosk:

```bash
sudo rm -rf /etc/systemd/system/getty@tty1.service.d
sudo systemctl daemon-reload
# rewrite ~/.bash_profile without the startx block
```

`.hushlogin` was tried first and did **not** fix it — it only suppresses the MOTD, not the agetty banner or the X.Org block.

## Scope note

[...]

PR #435, opened 2026-08-07
fix(kiosk): disable Openbox window decorations (#428)

nana-acorn:
Partial fix for #428 — removes the title bar. The white frame is still open there.

## The problem

At startup, before the Acorn splash paints, Chromium's window wears an Openbox title bar reading **"Untitled — Chromium"** across the top of the display.

`--kiosk` makes the window fullscreen and undecorated, but that state doesn't survive the window being re-mapped — it's re-applied *after* the window reappears, so there's a window of time where Openbox decorates it.

And nothing prevented that, because **we deployed no Openbox window rules at all** — no `rc.xml` anywhere in the repo, Phase 9b wrote only `autostart`, so Openbox ran stock defaults which decorate every window.

Worth noting the legacy WebKit kiosk guarded against this explicitly (`server_web/kiosk.py`, `set_decorated(False)`). The Chromium path relied on `--kiosk` alone and had no equivalent.

## The fix

A catch-all rule, since a kiosk never wants decoration on anything:

```xml
<!-- aquila-kiosk-no-decor: see #428 -->
<application class="*">
  <decor>no</decor>
</application>
```

**Verified on sn03** — title bar gone, kiosk otherwise unaffected.

## Three things about the implementation

**We now own a copy of `rc.xml`.** Openbox has no drop-in config directory; the only way to set a window rule is to own the whole file. That is precisely the "copied file drifts from upstream" pattern behind #424 and #426, both fixed this week. It's deliberate, and the copy carries a comment saying why it exists so the next person doesn't have to work it out.

**The result is XML-validated before it can reach a reboot.** A malformed `rc.xml` leaves Openbox unable to start — a blank screen on a device that may not be physically reachable. If validation fails the stock file is restored.

**The edit is idempotent.** Re-running deployment3 won't stack duplicate rules; a test enforces it.

## Tests

Five guard tests covering the rule, the copy-don't-hand-write choice, insertion inside the existing `<applications>` block (a second top-level one would be silently ignored), XML validation with rollback, and idempotency.

```
tests/fleet_device — 76 passed
bash -n scripts/deploy/deployment3_greengrass.sh — OK
```

The `sed` insertion was also run against a real `rc.xml` structure and the output confirmed to parse.

## Scope

Fixes the title bar only. Still open:

- **#428** — the white frame before the splash paints. Three theories ruled out and documented there (Chromium's blank page, the X root window, and hiding the window with `xdotool`). Plymouth is the proposed direction, since it's the one approach that doesn't require correctly identifying what the white is.
- **#431** — splash → app handoff, cross-origin renderer swap. Measured A/B on sn11.

`deployment2.sh` needs the same rule; it's a fork on `main`, out of scope here.

Fourth in a series with #419, #425 and #427, all off `feat/greengrass-migration`, all touching different phases.

### DIFF
--- a/scripts/deploy/deployment3_greengrass.sh
+++ b/scripts/deploy/deployment3_greengrass.sh
@@ -798,13 +798,21 @@ EOF
 # this explicitly (server_web/kiosk.py, set_decorated(False)); the Chromium path
 # relies on --kiosk alone and had no equivalent until now.
 #
+# The rule also forces the window fullscreen at map time. Without it Chromium's
+# window is mapped small and then resized up to fullscreen, and you watch it grow
+# — this is the "quadrant fill" reported on #428, which was mistaken for a slow
+# repaint for some time. Frame-by-frame capture on sn09 showed a part-sized,
+# decorated window rather than a partially painted one. Mapping it fullscreen
+# removes the intermediate sizes entirely.
+#
 # A kiosk never wants decoration on anything, so the rule is unconditional.
 # Openbox has no drop-in config directory — the only way to set a window rule is
 # to own a copy of rc.xml, which is why this copies the distro file rather than
 # patching it in place. That is the same "copied file drifts from upstream"
 # pattern that caused #424 and #426, so it is deliberate and documented here.
 #
-# Verified on sn03: title bar gone, kiosk otherwise unaffected.
+# Verified on sn03 (title bar gone) and sn09 (window arrives fullscreen; the
+# growing fill is gone). Kiosk otherwise unaffected on both.
 OPENBOX_RC="${PI_HOME}/.config/openbox/rc.xml"
 if [[ ! -f "${OPENBOX_RC}" && -f /etc/xdg/openbox/rc.xml ]]; then
     cp /etc/xdg/openbox/rc.xml "${OPENBOX_RC}"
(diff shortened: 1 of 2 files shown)


nana-acorn:
## Added: the browser window background (same surface, same problem)

Extended this PR to also fix the *colour* of the window this rule governs, since it turned out to be the same surface and the fix belongs in the same phase.

**The problem:** Chromium's window background — visible after the window is mapped but before the page paints — is bright white. Against the black either side of it in the startup sequence, that made every seam and flicker conspicuous.

**Why no Chromium flag fixed it:** it comes from GTK, not from Chromium. `--default-background-color` (tried as both `000000ff` and `ffffffff`) and `--cast-app-background-color` were each **verified present in the running process** on sn09 and changed nothing — they govern the page area, which is a different surface. That distinction cost a while to find.

**What does reach it:** a dark GTK theme. Confirmed on sn09 — the white immediately became dark grey and the tearing became far less visible. `GTK_THEME=Adwaita:dark` on the launch line selects it; a `gtk.css` override pins it to pure black so the window is indistinguishable from its surroundings.

### Honest scope

This removes the *contrast* that made the tearing visible, not the tearing. And there is a floor: the panel is briefly **unlit** during the display handover, and a lit black screen can never match an unlit one — that limit is documented in #428 along with the seven approaches ruled out against it.

### This PR now covers three things about the pre-splash window

| | Effect |
|---|---|
| `<decor>no</decor>` | no "Untitled — Chromium" title bar |
| `<fullscreen>yes</fullscreen>` | window arrives full-size instead of growing (the "quadrant fill") |
| dark GTK background | empty window matches its surroundings instead of flashing white |

All three verified on sn09, all three config-only with no boot risk, all three revertible by deleting a file.

PR #436, opened 2026-08-10
fix(kiosk): serve the kiosk front door from host nginx (#431)

### LINKED ISSUE #431: Splash-to-app handoff is glitchy: cross-origin redirect forces a second Chromium renderer

nana-acorn:
Closes #431.

The kiosk opens `http://localhost:8080/` and never navigates away. **Host nginx** answers that address: the boot splash while the backend is still starting, the app once it is up. Same origin either way.

## The problem

`splash.html` loaded from `file://` and redirected to `http://localhost:8090`. That crosses an origin boundary, so Chromium discards the renderer it just warmed up and builds a second one — the visible blank between splash and app on every boot.

**Measured A/B on sn11**, same page and flags, only the URL differing:

| Splash served from | Splash → app |
|---|---|
| same origin as the app | **seamless** |
| `file:///opt/aquila/splash.html` | visible gap |
| real boot | worst |

## Why host nginx, not the container

The first version of this PR pointed the kiosk at the existing `aquila-ui` container. **That does not work**, and the way it fails is instructive.

Measured on sn09:

```
18:00:14.011  aquila-stack.service — Starting
18:00:44.536  aquila-stack.service — Finished      ← ~30-40s
...
18:00:05.4    session opens
18:00:08.4    Chromium launches                    ← ~8s
```

Chromium asks roughly 36 seconds before the container is listening. Result: `ERR_CONNECTION_REFUSED` on screen, or — with a wait-for-nginx loop — a black screen for the entire 40 seconds with no splash at all. Both observed on sn09.

It passed on sn10 and sn11 because those boots won the race. The deciding factors are ordinary boot variance (SD speed, warm Docker layers, `wifi-recovery.service` at 10.4s), so **the failure is per-boot, not per-device** — the same unit could show the splash one morning and an error page the next, passing every test anyone happens to run.

Host nginx starts in under a second, before the desktop exists. There is no race to lose.

## What Phase 3d does

```nginx
location = / {
    proxy_pass http://127.0.0.1:8090;
    error_page 502 503 504 = @splash;
    proxy_connect_timeout 2s;
}

location @splash {
    try_files /splash.html =404;
    add_header Cache-Control "no-store, no-cache, must-revalidate" always;
}
```

- **Backend absent** → nginx cannot connect, serves `/opt/aquila/splash.html`.
- **Backend up** → proxies through on loopback, the app is served.

`no-store` matters: one URL serves two documents, so the transient one must never be cached, or the browser can re-serve the splash after the app is ready and sit there indefinitely. That was observed before the header was added.

`aquila-ui` moves to `:8082`. Both publishing 8080 would leave them fighting for the port and the stack failing to start.

## Verified end to end on sn09

```
$ curl -s http://localhost:8080/ | grep -o "<title>[^<]*</title>"
<title>Login</title>                  ← backend up: the app
$ docker stop aquila-backend
<title>Aquila — Starting</title>      ← backend down: the splash
$ docker start aquila-backend
<title>Login</title>                  ← back to the app
```

On a real boot: splash visible ~7 seconds **before the compose stack even starts**, then a seamless switch to the app. The generated config passes `nginx -t` (checked against a real nginx, not just asserted).

## Second problem this fixes

A `file://` page fetching `http://localhost:8090/health` is cross-origin, so the poll only worked because Phase 9b launched Chromium with **`--disable-web-security --allow-file-access-from-files`** — on a device holding device certificates and talking to AWS IoT.

Confirmed by accident: the `file://` splash without those flags hangs forever on the splash. Served same-origin the poll is simply allowed, and both flags are gone.

**This requires the relative-URL `splash.html` in the same change.** Ship one without the other and the device sits on the splash forever — observed on sn09 when the old absolute-URL file was still in place.

## Also in this PR — `docker/nginx.conf` correctness fixes

Independently valuable regardless of how the splash is served:

- **Per-request upstream resolution.** A literal upstream name is resolved once at startup and nginx refuses to start if the backend container is absent — the failure that took down aquila-ui on sn01. The existing test guarded this for `host.docker.internal` only; the backend upstream had the same latent bug.
- **`resolver_timeout`.** With nginx's defaults a missing backend stalled every health poll for a full minute (sn10: 504 in 60.03s).
- **`/static/` alias.** `Dockerfile.ui` copies `aquila_web/static/` *into* the document root, so the files are flat and a root+try_files lookup 404s — the app rendered unstyled on sn10.
- **`no-store` on the splash fallback.**

One existing test was updated rather than deleted: `test_backend_upstream_uses_container_name` asserted a literal `proxy_pass` string. The requirement — container name, not localhost or an IP — is unchanged; only where the name is written moved into a `set` directive. It now checks intent and additionally rejects `localhost`/`127.0.0.1` upstreams.

A bug in `_location_blocks()` was also fixed: it reset brace depth after consuming the opening `{`, truncating every block to one body line, so several guard tests were inspecting almost nothing and passing by luck.

## Tests

```
tests/fleet_device — 85 passed
bash -n scripts/deploy/deployment3_greengrass.sh — OK
generated nginx config — nginx -t successful
```

## Open questions for review

- **`/kiosk-control/`** was proxied by the container's config and is not in the host config. If anything uses that path it needs adding.
- **Is `aquila-ui` now redundant?** It still serves `/ready`, `/run`, `/complete` and `version.json`. If those can move, the fleet ends up with one nginx instead of two — a simplification rather than the duplication this currently introduces. Worth deciding before this lands widely.
- **The 30-40s stack startup is untouched.** Host nginx makes that wait *visible* (splash instead of black) rather than shorter. It remains the largest single thing in the boot, by far, and deserves its own issue.

### DIFF
--- a/aquila_web/static/splash.html
+++ b/aquila_web/static/splash.html
@@ -124,13 +124,13 @@
     const MAX_WAIT = 120;
 
     function poll() {
-      fetch('http://localhost:8090/health', { cache: 'no-store' })
+      fetch('/health', { cache: 'no-store' })
         .then(r => {
           if (r.ok) {
             statusEl.textContent = 'Ready';
             bar.style.transition = 'width 0.4s ease';
             setBar(100);
-            setTimeout(() => { window.location.href = 'http://localhost:8090'; }, 450);
+            setTimeout(() => { window.location.href = '/'; }, 450);
           } else { retry(); }
         })
         .catch(() => retry());
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -5,8 +5,35 @@ server {
     root /usr/share/nginx/html;
     index index.html;
 
+    # Docker's embedded DNS. Every proxy_pass below resolves through a variable
+    # so the lookup happens per request, not at startup: nginx must come up and
+    # stay up while aquila-backend does not exist yet, which is the whole point
+    # of the boot splash. A literal upstream name is resolved once at startup
+    # and takes nginx down with it — the failure that killed aquila-ui on sn01
+    # (see tests/fleet_device/test_nginx_config.py).
+    resolver 127.0.0.11 valid=10s ipv6=off;
+
+    # The kiosk opens this address and never navigates away from it (#431).
+    # Backend up   -> proxied through, the app is served.
+    # Backend down -> nginx cannot connect, and @splash serves the boot splash.
+    #
+    # Same origin either way, so handing off from splash to app reuses Chromium's
+    # warm renderer instead of swapping processes. Measured on sn11: served from
+    # this origin the handoff is seamless; from file:// it visibly blanks.
     location = / {
-        try_files /index.html =404;
+        set $aq_backend http://aquila-backend:8090;
+        proxy_pass         $aq_backend;
+        error_page         502 503 504 = @splash;
+        proxy_http_version 1.1;
+        proxy_set_header   Host $host;
+        proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_connect_timeout 2s;
+    }
+
+    # Served from the UI image's own copy of aquila_web/static/ — the same file
+    # the repo ships. Nothing needs to install it on the host.
+    location @splash {
+        try_files /splash.html =404;
     }
 
     location = /ready {
@@ -45,8 +72,13 @@ server {
         proxy_connect_timeout 5s;
     }
 
+    # Everything else — API calls, /health, websockets — goes to the backend.
+    # Variable upstream for the same reason as above: a literal name here meant
+    # nginx refused to start whenever the backend container was absent, so the
+    # UI died alongside the thing it was meant to survive.
     location / {
-        proxy_pass http://aquila-backend:8090;
+        set $aq_backend_all http://aquila-backend:8090;
+        proxy_pass $aq_backend_all;
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -12,6 +12,12 @@ server {
     # and takes nginx down with it — the failure that killed aquila-ui on sn01
     # (see tests/fleet_device/test_nginx_config.py).
     resolver 127.0.0.11 valid=10s ipv6=off;
+    # While the backend container does not exist its name does not resolve. The
+    # default 30s resolver timeout and 60s proxy timeouts made every splash
+    # health poll hang for a full minute — measured on sn10: 504 in 60.03s — so
+    # the splash sat there long past the app being ready. Fail fast instead, and
+    # let the page poll at the once-a-second rate it was written for.
+    resolver_timeout 2s;
 
     # The kiosk opens this address and never navigates away from it (#431).
     # Backend up   -> proxied through, the app is served.
@@ -48,8 +54,16 @@ server {
         try_files /complete.html =404;
     }
 
+    # The app's HTML references its assets relatively (href="static/styles.css"),
+    # which resolves to /static/… — but Dockerfile.ui copies aquila_web/static/
+    # INTO the document root, so the files sit at the top level and there is no
+    # static/ subdirectory. `alias` maps the URL prefix onto the flat directory.
+    # try_files is deliberately absent: combined with alias it resolves $uri
+    # against the aliased path and yields /usr/share/nginx/html/static/… again.
+    # Measured on sn10 before this: /static/styles.css returned 404 and the app
+    # rendered unstyled.
     location /static/ {
-        try_files $uri =404;
+        alias /usr/share/nginx/html/;
     }
 
     # The UI image's own baked identity (#351). This MUST be served locally: the
@@ -79,6 +93,11 @@ server {
     location / {
         set $aq_backend_all http://aquila-backend:8090;
         proxy_pass $aq_backend_all;
+        # Bounded so /health fails fast while the backend is still starting —
+        # the splash polls once a second and must not be blocked for 60s.
+        proxy_connect_timeout 2s;
+        proxy_send_timeout    10s;
+        proxy_read_timeout    60s;
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -40,6 +40,13 @@ server {
     # the repo ships. Nothing needs to install it on the host.
     location @splash {
         try_files /splash.html =404;
+        # / serves two different documents depending on whether the backend is
+        # up, so the transient one must never be cached. Without this the browser
+        # is entitled to re-serve the splash from cache when the page navigates
+        # back to / after the app is ready, and the kiosk sits on the splash long
+        # after it should have moved on.
+        add_header Cache-Control "no-store, no-cache, must-revalidate" always;
+        expires -1;
     }
 
     location = /ready {
--- a/fleet-config/greengrass-compose.yaml
+++ b/fleet-config/greengrass-compose.yaml
@@ -135,7 +135,11 @@ services:
     env_file: ${DEVICE_ENV_FILE:-/opt/aquila/config/device.env}
     container_name: aquila-ui
     ports:
-      - "8080:80"
+      # 8080 belongs to host nginx, which is the kiosk's front door (#431):
+      # it must answer within a second of boot, and this container does not
+      # exist for the first ~40s. Publishing both here would leave the two
+      # fighting for the port and the stack failing to start.
+      - "8082:80"
     extra_hosts:
       - "host.docker.internal:host-gateway"
     restart: unless-stopped
--- a/scripts/deploy/deployment3_greengrass.sh
+++ b/scripts/deploy/deployment3_greengrass.sh
@@ -292,6 +292,17 @@ rm -f /etc/nginx/sites-enabled/default
 # new config would never load. Restart explicitly. (Cost an hour on sn09.)
 nginx -t
 systemctl enable nginx
+
+# On a re-run, a previous deployment's aquila-ui may still be publishing :8080
+# and nginx cannot bind — `bind() to 0.0.0.0:8080 failed (98: Address already in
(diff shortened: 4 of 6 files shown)


nana-acorn:
## ⚠️ Blocked — do not merge as written

Testing on sn09 found a flaw this PR's verification missed. The measured handoff improvement is real, but the change gives away something more valuable.

### The splash can no longer appear until Docker is up

nginx runs in Docker. Measured on sn09:

```
17:54:48.363  aquila-stack.service — Starting
17:55:28.815  aquila-stack.service — Finished     ← ~40 seconds
```

`systemd-analyze blame`: `docker.service` alone takes 11.96s.

The old `file://` splash needed nothing and was on screen the moment Chromium launched. Pointed at `http://localhost:8080/`, the kiosk has nothing to show until nginx answers — so that wait moves **from the splash to a black screen**.

Observed on sn09 with a wait-for-nginx loop added: the splash never appeared at all. By the time nginx answered, the backend was up too, so nginx proxied straight to the app — after a long black screen. Without the loop, the kiosk gets `ERR_CONNECTION_REFUSED` and shows Chromium's "This site can't be reached" page instead.

Neither is acceptable. The splash exists precisely to cover that wait.

### The original verification passed by luck

This PR was verified on sn10 and sn11, where nginx happened to be answering before Chromium asked. There was no wait loop — it was a race that those boots won.

sn09 loses it. And the deciding factors are ordinary boot variance (SD speed, warm Docker layers, `wifi-recovery.service` at 10.4s, `NetworkManager-wait-online` at 6.8s), so **this is per-boot, not per-device**. The same unit could show the splash one morning and an error page the next — passing every test anyone happens to run.

The risk was named in the issue ("Needs measuring that nginx is actually up before Chromium launches") and never measured. That was the gap.

### What would actually work

Run nginx **on the host as a systemd service** rather than in Docker. It starts in under a second, serves the splash immediately, and proxies to the backend once it appears — same origin, no wait, no race. That keeps everything this PR set out to achieve, including dropping `--disable-web-security`.

That is a materially bigger change than this PR describes and deserves its own design, not a patch on top of this one.

### Still valid from this work

- The A/B measurement stands: same-origin genuinely removes the handoff glitch.
- `--disable-web-security` and `--allow-file-access-from-files` are only needed by the `file://` path and should go whenever this lands in some form.
- The `nginx.conf` fixes here are independently correct and worth keeping regardless of how the splash is served: per-request upstream resolution (a literal name takes nginx down when the backend is absent — the sn01 failure), `resolver_timeout` (a missing backend stalled every health poll for 60s), the `/static/` alias (the app rendered unstyled), and `no-store` on the splash fallback.

Suggest splitting: land the `nginx.conf` correctness fixes, and re-do the kiosk-origin change on top of a host-side nginx.

nana-acorn:
## ✅ Unblocked — reworked to host nginx

The blocking objection above is resolved. Summary for anyone reading the thread in order:

**What was wrong:** the kiosk pointed at the `aquila-ui` container on :8080, which does not exist for the first ~30-40s of boot. Chromium asks at ~8s. Result was `ERR_CONNECTION_REFUSED`, or with a wait loop, 40s of black and no splash. It passed initial testing on sn10/sn11 only because those boots won the race — the failure was per-boot, not per-device.

**What changed:** new Phase 3d installs nginx **on the host**. It starts in under a second, so it is answering long before Chromium asks. Splash while the backend is absent, proxy to `127.0.0.1:8090` once it is up.

**Verified on sn09** end to end, including the boot: splash visible ~7s before the compose stack even starts, then a seamless switch to the app. Both states confirmed by `curl` with the backend stopped and running. Generated config passes `nginx -t`.

**Also fixed along the way**, each found by it breaking on a device rather than by review:

- The splash must be the relative-URL version, or a same-origin front door leaves the device stuck on the splash forever (the absolute `localhost:8090` URLs are cross-origin from :8080, and `--disable-web-security` is now gone). Asserted by a test.
- nginx must be `restart`ed, not just `enable --now`'d — the package starts it before the config exists, and `--now` does nothing to a running service. Asserted by a test.
- `aquila-ui` moves to :8082, or the two fight for the port and the stack fails to start. Asserted by a test.

**Two questions I would like a reviewer's view on** rather than deciding alone, both in the PR body: whether `/kiosk-control/` needs adding to the host config, and whether `aquila-ui` is now redundant — if it is, this becomes a simplification rather than a second nginx to maintain.

nana-acorn:
## Concern: this leaves two nginx configs, and they will drift

Worth stating plainly, because it is the same failure mode this epic exists to clean up.

After this PR the fleet has **two** nginx configurations:

| | Where | Serves |
|---|---|---|
| `docker/nginx.conf` | `aquila-ui` container, now :8082 | `/ready`, `/run`, `/complete`, `/static/`, `/version.json`, `/kiosk-control/`, catch-all proxy |
| `/etc/nginx/conf.d/aquila-kiosk.conf` | host, :8080 | `/` with splash fallback, catch-all proxy |

Both proxy to the same backend. Both will need the same fixes when something changes. Neither knows about the other.

**#424 and #426 — both fixed in this epic — were caused by exactly this**: a config that outlived its purpose, drifted from the thing that replaced it, and eventually produced a customer-visible fault years later. Adding a second copy deliberately, having just spent a week removing the last one, deserves scepticism rather than a shrug.

I would rather flag it now than have it discovered the same way the others were.

## Can they be merged? Probably — and it may be a net deletion

The container's unique responsibilities look thinner than they first appear:

**Static pages and assets.** The backend already serves these. Verified on sn09: `http://localhost:8090/` returns the app and `http://localhost:8090/static/styles.css` returns 200. `Dockerfile.ui` copies the *same* `aquila_web/static/` directory the backend serves from, so the container is a local cache of files the backend already has.

**`/kiosk-control/`.** The container proxies to `host.docker.internal:9191` and needs a `resolver` and a variable upstream to avoid crashing at startup (the sn01 failure). **From the host that is just `127.0.0.1:9191`** — no resolver, no variable, no crash mode. It gets simpler, not harder.

**`/version.json` (#351).** The genuine one. It exists so the UI image reports *its own* build identity rather than the API container's, which is the whole point of that check. If the UI image goes away, that check needs rethinking — it cannot simply move to the host, or it stops measuring what it was built to measure.

So the shape of a consolidation is:

1. Host nginx takes `/kiosk-control/` (simpler there than in the container).
2. Host nginx proxies everything else to the backend, which already serves it.
3. `aquila-ui` is retired, `docker/nginx.conf` and `Dockerfile.ui` are deleted.
4. **#351's image-identity check needs a new home** — this is the blocker and the part I cannot resolve without knowing what that check is for in practice.

That ends with **one** nginx config instead of two, one fewer container, and one fewer image to build and pull on every device. A simplification.

## What I would suggest

Land this PR — the front door has to be on the host regardless, and the measured failure it fixes is real.

Then open a follow-up for the consolidation, with #351 as the open question. Doing it in this PR would mean deciding the `version.json` question in passing, and that deserves its own discussion with whoever added it.

If the consolidation is not going to happen, that is a reasonable answer too — but then the two configs should at least reference each other in comments, so the next person changing one knows the other exists.

nana-acorn:
Consolidation tracked separately in #444 — retire `aquila-ui`, move `/kiosk-control/` to the host config, and rehome the #351 `version.json` check, which is the blocker. That would end with one nginx instead of two and one fewer image to build and pull.

Not a prerequisite for this PR: the front door has to be on the host regardless, and the failure it fixes is measured. But the two-config state this leaves behind is the same pattern as #424 and #426, so it should not sit unexamined.

PR #438, opened 2026-08-11
fix(display): rotate at X startup; fix the logo jump via Chromium's initial window size (#437)

### LINKED ISSUE #437: Display rotation is applied from the session, so the splash re-layouts on screen at every boot
## Summary

The panel is portrait but the display comes up landscape, and the rotation is applied by `xrandr --rotate right` from the **Openbox autostart** — i.e. from inside the desktop session, after X, LightDM and Chromium have already started painting.

Two visible defects follow from that, both on every boot:

1. **The Acorn logo appears high on the screen and visibly drops into place.** The splash centres itself in the viewport; Chromium lays it out against a landscape viewport (768 tall), then re-lays it out when the geometry becomes portrait (1024 tall), so the logo is drawn high and then moves down.

2. **The screen fills in blocks rather than as one frame** — top-left, then top-right, then the lower half — while the rotated framebuffer is repainted region by region.

Customer-facing: this is the first thing on screen at every power-on.

## Evidence

Captured frame-by-frame on **sn10**. Before the change: partial quadrant fill, then the splash with the logo positioned high, then the logo settling to centre.

Current state on a device:

```
$ xrandr --query | grep " connected"
HDMI-2 connected 768x1024+0+0 right (normal left inverted right x axis y axis) 162mm x 122mm
```

`right` is present, but only because the session applied it seconds after boot.

## Fix

Rotate at X startup instead, via `/etc/X11/xorg.conf.d/99-rotate.conf`, and drop `--rotate` from the session's `xrandr` call. X is then portrait from its first frame and nothing re-layouts.

**Verified on sn10:** the logo jump is gone and the block fill is much reduced (now roughly half-screen rather than quadrant-by-quadrant, and faster). Touch remains correctly mapped.

Both connector names are written. The panel is not always on the same one — sn10 reports HDMI-2, and the autostart has always had to auto-detect it. Xorg ignores a Monitor section whose Identifier matches no output, so the unused section is harmless.

## What this does not fix

Two flashes remain in the pre-splash sequence, both visible in the captured frames:

- The block fill is reduced but not gone. If that is the compositor rotating each frame, only rotating at the kernel level (`video=HDMI-A-2:…,rotate=90` in `cmdline.txt`) would remove it — a larger change, untested.
- A brief black frame between X finishing its paint and Chromium's window appearing. That is the X/LightDM → Chromium handover, and belongs to #428.

## Related

- **#428** — white frame before the splash paints. Adjacent but distinct; that issue's remaining work is Plymouth.
- **#431** — splash → app handoff. Fixed separately.
- `config.txt` sets `display_hdmi_rotate=1` *and* the autostart rotated again. The `config.txt` setting is inert under the KMS driver, which is presumably why the `xrandr` workaround exists. Worth cleaning up once rotation has a single home.

nana-acorn:
Closes #437. Two separate things, both about the boot display — and the second one corrects what this PR originally claimed.

## 1. Rotate the panel at X startup, not from the session

The panel is portrait, the display comes up landscape, and rotation was applied by `xrandr --rotate right` from the **Openbox autostart** — from inside the session, after X, LightDM and Chromium had started painting.

Rotating in `/etc/X11/xorg.conf.d/99-rotate.conf` means X is portrait from its first frame. Confirmed applied on sn09:

```
$ grep -i rotate /var/log/Xorg.0.log
[    20.294] (**) modeset(0): Option "Rotate" "right"
```

Both connector names are written — the panel is not always on the same one (sn09 and sn10 report HDMI-2), which is why the autostart has always had to auto-detect it. Xorg ignores a Monitor section matching no output.

Touch was verified after the change; the autostart's `Coordinate Transformation Matrix` still maps correctly.

## 2. The logo jump — and a correction

**This PR originally claimed late rotation caused the logo jump. It does not.** Rotation was confirmed applied at the X level and the jump still occurred.

Frame-by-frame geometry logging on sn09 found the real cause. Every boot showed:

```
10x10     ← Chromium's internal window
748x561   ← the kiosk window, mapped small, held ~1s
768x1024  ← finally fullscreen
```

| Boot | at 748×561 | → 768×1024 | duration |
|---|---|---|---|
| 14:48 | 14:49:09.7 | 14:49:10.8 | 1.09s |
| 15:38 | 15:41:02.5 | 15:41:03.6 | 1.07s |
| 15:41 | 15:42:34.4 | 15:42:35.7 | 1.26s |
| 15:47 | 15:48:52.2 | 15:48:53.1 | 0.87s |
| 15:50 | 15:51:54.8 | 15:51:55.8 | 0.97s |

The splash centres itself in the viewport, so a page painting during that second is laid out for a 561px-tall window and its content drops when the window grows to 1024.

**This also explains the intermittency**, which nothing else did. The small window happens on *every* boot; the jump only appears when the page paints before the resize rather than after. That is why the symptom came and went on identical configurations and why single-boot observations were worthless.

`--window-size=768,1024 --window-position=0,0` reduced the intermediate stage to `767x1023` across **eight consecutive boots** — a one-pixel change instead of a 463-pixel one.

**A residual ~70px shift remained**, inside the page rather than the window. Centring positions the block by its total height, so anything resizing after first paint moves the logo. Anchoring from the top with `padding-top` makes the position depend only on the viewport, so no late reflow can move it. That fixes the class of problem without identifying which element resized — several candidates were checked and none confirmed.

## Why the rotation change stays

Applying rotation before anything paints is correct regardless of the jump, and the previous ordering was a genuine hazard — the session rotating after Chromium had painted is the kind of thing that produces exactly this class of bug. It just was not the cause here.

## Not isolated

Whether the rotation change *also* contributes to the jump is unknown: both were applied throughout testing on sn09, and the isolation run was not completed. This PR claims only what the geometry data shows.

## Verification

- Rotation applied at X level, confirmed in `Xorg.0.log`, touch correct.
- `--window-size`: eight consecutive boots, no `748x561` stage.
- Splash anchoring: verified visually on sn09; position adjusted to 25% to match the previous centred placement.
- `bash -n` clean, `tests/fleet_device` — 110 passed.

One thing worth flagging for reviewers: an earlier revision of this commit placed the explanatory comment *inside* the backslash-continued `chromium` invocation. `bash -n` passes on that, but it silently truncates the command and turns the following flags into a separate command. The comment now sits above the launch. Worth watching for in review — a passing syntax check does not catch it.

### DIFF
--- a/aquila_web/static/splash.html
+++ b/aquila_web/static/splash.html
@@ -13,7 +13,13 @@
       display: flex;
       flex-direction: column;
       align-items: center;
-      justify-content: center;
+      /* Anchored from the top rather than centred: centring positions this
+         block by its TOTAL height, so anything that resizes after first paint
+         (a font resolving, an element getting its real dimensions) moves the
+         logo. Measured on sn09 as a ~70px shift shortly after load. Anchoring
+         makes the logo's position depend only on the viewport. */
+      justify-content: flex-start;
+      padding-top: 25%;
       font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", sans-serif;
       overflow: hidden;
     }
--- a/scripts/deploy/deployment3_greengrass.sh
+++ b/scripts/deploy/deployment3_greengrass.sh
@@ -297,11 +297,43 @@ phase_pass "LightDM configured for X11/Openbox autologin (Wayland compositor dis
 # ═══════════════════════════════════════════════════════════════════════════════
 phase_start 6 "Display and Touch configuration"
 
+# Rotate the panel at X startup rather than from the Openbox autostart.
+#
+# The panel is portrait but the display comes up landscape, and rotation used to
+# be applied by `xrandr --rotate right` from the session — after X, LightDM and
+# Chromium had already painted. Chromium therefore laid the splash out against a
+# landscape viewport, then re-laid it out when the geometry changed: measured on
+# sn10, the Acorn logo appeared high on the screen and visibly dropped into place.
+# The screen also filled in blocks rather than as one frame while the rotated
+# framebuffer was repainted.
+#
+# Setting it here means X is portrait from its first frame, so nothing re-layouts.
+# Verified on sn10: the logo jump is gone and the block fill is much reduced.
+#
+# Both connector names are written because the panel is not always on the same
+# one — sn10 reports HDMI-2 (kernel HDMI-A-2), and the autostart has always had to
+# auto-detect it. Xorg ignores a Monitor section whose Identifier matches no
+# output, so the unused one is harmless.
+mkdir -p /etc/X11/xorg.conf.d
+cat > /etc/X11/xorg.conf.d/99-rotate.conf <<'EOF'
+Section "Monitor"
+    Identifier "HDMI-1"
+    Option "Rotate" "right"
+EndSection
+
+Section "Monitor"
+    Identifier "HDMI-2"
+    Option "Rotate" "right"
+EndSection
+EOF
+
 # xrandr and xinput are called at runtime from Openbox autostart (Phase 9b).
 # This phase verifies the required tools are present on the host.
 run_test "xrandr binary present"  "which xrandr"
 run_test "xinput binary present"  "which xinput"
 run_test "unclutter present"      "which unclutter"
+run_test "X rotation config"      "test -f /etc/X11/xorg.conf.d/99-rotate.conf"
+run_test "rotation set to right"  "grep -q 'Option \"Rotate\" \"right\"' /etc/X11/xorg.conf.d/99-rotate.conf"
 
 # Disable the Raspberry Pi welcome wizard so it never appears on first boot
 rm -f /etc/xdg/autostart/piwiz.desktop
@@ -746,7 +778,10 @@ xset -dpms
 # Auto-detect connected HDMI output (handles HDMI-2, HDMI-A-2, etc.)
 HDMI_OUT=$(xrandr --query | grep -E "^HDMI.* connected" | head -1 | awk '{print $1}')
 if [ -n "$HDMI_OUT" ]; then
-    xrandr --output "$HDMI_OUT" --mode 1024x768 --rate 60 --rotate right
+    # No --rotate here: X is already portrait from /etc/X11/xorg.conf.d/99-rotate.conf
+    # (Phase 6). Rotating again in the session would undo it, and rotating late is
+    # what made the splash re-layout on screen in the first place.
+    xrandr --output "$HDMI_OUT" --mode 1024x768 --rate 60
 fi
 
 xinput set-prop "Focaltech Systems FT5926 MultiTouch" \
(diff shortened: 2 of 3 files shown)


nana-acorn:
## ⚠️ Evidence correction — read before reviewing

The verification in this PR is weaker than the description implies. Further testing on sn10 showed **the logo jump is intermittent, not deterministic**, which undermines a single-boot before/after comparison.

### What was actually observed

| Configuration | Boots | Logo jump |
|---|---|---|
| Before the change | 1 | present |
| After the change | 1 | absent |
| **Fully reverted** (no change from this epic) | 1 | **absent** |
| **Fully reverted**, later | 1 | **present** |

The last two rows are the problem: the same unmodified configuration produced both outcomes. So "absent after the change" is consistent with the fix working *and* with simply having won the race that boot.

### What this does and doesn't change

**Still established:** the mechanism. Rotation was applied from the session, after Chromium had painted, and Chromium demonstrably re-laid the splash out when the geometry changed. Rotating in X before anything paints removes that ordering hazard rather than relying on winning it. That argument stands on its own.

**Still established:** the jump is pre-existing. It occurs with none of this epic applied, so it is not a regression we introduced — an earlier single observation suggested otherwise and was wrong.

**No longer established:** that this change eliminates it. One clean boot is not evidence against an intermittent symptom.

### What would settle it

Count rather than observe: 10 boots reverted, 10 boots with this applied, and compare rates. A baseline of ~4/10 dropping to 0/10 would be real evidence; 3/10 would mean this fixes nothing.

### Related, and possibly the same root

The same testing found an **intermittent screen flicker** on a fully-reverted sn10 — during the splash and occasionally in the app. Removing the `sleep 3` before the kiosk launch made it appear in 2 of 5 boots, which points at Chromium starting before the display has settled. That `sleep` is a fixed-time guess, and it is demonstrably marginal under load.

Both symptoms may share that root cause. Being filed separately; it is pre-existing and unrelated to this epic.

### Recommendation

Merge on the mechanism if reviewers agree rotating before first paint is correct regardless — but not on the strength of the verification as originally written.

nana-acorn:
## Correction — the earlier evidence note is superseded

The comment above said the verification was single-boot on an intermittent symptom and that a fix could not be told from a lucky boot. That was right at the time, and the counting it asked for has now been done — with a different answer than expected.

**The logo jump was never caused by late rotation.** Rotation was confirmed applied at the X level (`(**) modeset(0): Option "Rotate" "right"`) and the jump still occurred at roughly 1-in-5.

Frame-by-frame geometry logging found it: Chromium maps its window at **748×561** and holds it for about a second before the WM fullscreens it. The splash centres itself in the viewport, so a page painting during that second is laid out for a 561px-tall window and drops when it grows to 1024.

That also resolves the intermittency the earlier note flagged as unmeasured. The small window happens on **every** boot — the jump only shows when the page paints before the resize. So it was never a question of whether the fix worked, but of which side of a race the paint landed on.

The rotation change stays: rotating before anything paints is correct on its own merits, and the previous ordering was a real hazard. The PR title and description now say that rather than claiming it fixes the jump.

**Still not isolated:** whether rotation also contributes. Both changes were live throughout testing on sn09 and the isolation run was not completed. The PR says so explicitly rather than implying the rotation change is unnecessary.

